PR #31, opened 2015-01-26
Support email verification ticket

sandrinodimattia:
Added GenerateVerificationTicket method to the client to allow generation of the email verification url. This is useful when people send out custom emails in which they want to include an email verification link.


siacomuzzi:
Available @ [Auth0 v1.10.2](https://www.nuget.org/packages/Auth0)

Thanks @sandrinodimattia!

PR #32, opened 2015-02-10
Client.cs

edualv:
added validation for domains that already contains https


siacomuzzi:
Thanks for contributing to this!

Instead of use `string.Contains()`, could you please validate the domain url? Something like the following:

```
this.client = new RestClient(Uri.IsWellFormedUriString(this.domain, UriKind.Absolute) ? this.domain : "https://" + this.domain);
```

siacomuzzi:
Available @ `v1.10.3`

Thanks!

PR #36, opened 2015-02-27
Support username/password validation

sandrinodimattia:
This endpoint is required if you're building a secure password reset form which doesn't use email verification. You could present the user with 4 input fields:

 - Username
 - Current password
 - New password
 - Confirm new password

The username and current password would then be validated, and if both are correct the user's password can be changed to the new password.


dschenkelman:
Nice, didn't even knew that endpoints exists.

PR #37, opened 2015-04-09
Consistent, simplified and improved error messages

barclayadam:
I have been having issues with GetErrorDetails throwing a NullReferenceException when no details exist in a response (using an older client version). 

Although a fix was implemented about a month ago this pull request improves on that change by avoiding exceptions altogether plus a few other enhancements such as avoiding a JSON round-trip and providing a simple default message in the case of no further details.

### DIFF
--- a/Auth0.Net/Client.cs
+++ b/Auth0.Net/Client.cs
@@ -181,9 +181,7 @@ namespace Auth0
             var result = this.client.Execute(request);
             if (result.StatusCode != HttpStatusCode.OK && result.StatusCode != HttpStatusCode.Created)
             {
-                var detail = GetErrorDetails(result.Content);
-                throw new InvalidOperationException(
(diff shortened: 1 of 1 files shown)


Auth0.Net/Client.cs
```diff
@@ -1115,20 +1090,11 @@ public UserValidationResult ValidateUser(string username, string password, strin
...
-            {
-                return JsonConvert.DeserializeObject(resultContent).ToString();
-            }
-            catch (NullReferenceException)
-            {
-            }
-            catch (JsonReaderException)
-            {
-            }
+            var detail = response.Content ?? "No error details available";
```
siacomuzzi: what if `response.Content` is a json?

siacomuzzi:
Available @ `v1.11.0`

Thanks for contributing!

PR #43, opened 2015-07-15
Improve error handling in ExchangeAuthorizationCodePerAccessToken

twistedstream:
Under certain circumstances, the RestSharp response has no body and is
causing a `NullReferenceException`.  This change prevents that as well as reports back more meaningful errors for other scenarios.

### DIFF
--- a/Auth0.Net/Client.cs
+++ b/Auth0.Net/Client.cs
@@ -332,9 +332,9 @@
             var body = response.Data;
             if (response.StatusCode != HttpStatusCode.OK)
             {
-                var errorDescription = body.ContainsKey("error_description") ? body["error_description"] : "(no description)";
+                var errorDescription = (body != null && body.ContainsKey("error_description")) ? body["error_description"] : "(no description)";
                 errorDescription += "; StatusCode=" + (int)response.StatusCode;
-                var errorCode = body.ContainsKey("error") ? body["error"] : string.Empty;
+                var errorCode = (body != null && body.ContainsKey("error")) ? body["error"] : string.Empty;
 
                 throw new OAuthException(errorDescription, errorCode);
             }
--- a/Auth0.Net/Client.cs
+++ b/Auth0.Net/Client.cs
@@ -321,7 +321,6 @@
             request.AddParameter("redirect_uri", redirectUri, ParameterType.GetOrPost);
 
             var response = this.client.Execute<Dictionary<string, string>>(request);
-            var body = response.Data;
 
             if (response.ResponseStatus != ResponseStatus.Completed)
             {
@@ -330,6 +329,7 @@
                     string.Empty);
             }
 
+            var body = response.Data;
             if (response.StatusCode != HttpStatusCode.OK)
             {
                 var errorDescription = body.ContainsKey("error_description") ? body["error_description"] : "(no description)";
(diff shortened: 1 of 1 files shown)


Auth0.Net/Client.cs
```diff
@@ -320,17 +320,29 @@ public TokenResult ExchangeAuthorizationCodePerAccessToken(string code, string r
             request.AddParameter("grant_type", "authorization_code", ParameterType.GetOrPost);
             request.AddParameter("redirect_uri", redirectUri, ParameterType.GetOrPost);
 
-            var response = this.client.Execute<Dictionary<string, string>>(request).Data;
+            var response = this.client.Execute<Dictionary<string, string>>(request);
+            var body = response.Data;
```
ntotten: This seems odd here. It might be better to access `response.Data` after you do all the checks to make sure the response is valid.

Auth0.Net/Client.cs
```diff
@@ -320,17 +320,29 @@ public TokenResult ExchangeAuthorizationCodePerAccessToken(string code, string r
...
+            }
+
+            if (response.StatusCode != HttpStatusCode.OK)
+            {
+                var errorDescription = body.ContainsKey("error_description") ? body["error_description"] : "(no description)";
+                errorDescription += "; StatusCode=" + (int)response.StatusCode;
+                var errorCode = body.ContainsKey("error") ? body["error"] : string.Empty;
+
+                throw new OAuthException(errorDescription, errorCode);
             }
```
ntotten: It might be worth checking `response.Data` here. Not sure if it can be null or not, but if it can it would be better to throw here than when `body.ContainsKey` fails.

```cs
var body = response.Data;
if (body == null) { 
  throw AppropriateExcecption('blah blah');
}
```

twistedstream:
There are valid scenarios where body == null.  However, I have now moved the creation of the `body` variable down to just before it is actually used.

ntotten:
But if `body == null` it is going to throw a NullReferenceExcpetion on the line with `AccessToken = body["access_token"],` That doesn't seem like a very useful exception. What are the conditions that body is null? Shouldn't you throw an exception stating said conditions?

twistedstream:
Body may be null if StatusCode != OK (on [these lines](https://github.com/auth0/auth0.net/blob/better-error-handling-in-ExchangeAuthorizationCodePerAccessToken/Auth0.Net/Client.cs#L333-L340)) but it should not be null after that.  We could still do a null check just to be super thorough but that would mean there's something wrong with the actual responses coming back from the Auth0 API.  If you'd like that, I can add it.  Not sure what the right exception class should be since its not really the fault of the app.

And FYI: the previous version of the SDK never checked for null in any circumstance.

twistedstream:
@ntotten, actually since `body` could be null when `StatusCode != OK` I needed to add null checks in that block, [which I've now done](https://github.com/auth0/auth0.net/blob/better-error-handling-in-ExchangeAuthorizationCodePerAccessToken/Auth0.Net/Client.cs#L335-L337).

So now its your call whether we really need to do the same in the [final code block](https://github.com/auth0/auth0.net/blob/better-error-handling-in-ExchangeAuthorizationCodePerAccessToken/Auth0.Net/Client.cs#L342-L346) since the Auth0 API should always return a body in this case.

PR #46, opened 2015-11-04
Exchage error information extended

dthpth:
RestResponse error exception added into OAuthException as inner
exception


ntotten:
Complete rewrite coming soon in v2 branch.

PR #48, opened 2015-11-18
Initial work on V2

jerriep:
This is the initial work on V2. I want to get it in the main repo so people can start commenting on it.


advancedrei:
So, quick observation. The name `Auth0.Api.Management` seems like it might be related to managing the API or something. Also, because this is not the API framework itself, just the client to access it, I think "Client" needs to be in the name. Since this is the API Client for managing your account, I think it needs to be named "Auth0.ManagementApiClient", or `Auth0.ManagementApi.Client` if you guys ever plan to ship the server-side API code, and the more limited version needs to be `Auth0.ApiClient` or `Auth0.Api.Client` or something very similar.

jerriep:
Yeah, so I think the main thing is to differentiate between the 2 APIs - the Management API, and the Authentication API:

image

How about we go with `Auth0.ManagementApi.Client` and `Auth0.AuthenticationApi.Client` respectively?

@advancedrei I am not sure I am on the same page when you talk about a "limited version"... Are you talking about the Authentication API or is this something else you are talking about?

advancedrei:
I'm fine with those names. The "limited" version was the Authentication version. That client is not going to do much compared to the Management one.

ntotten:
I'm fine with whatever you guys decide. :)

PR #53, opened 2015-11-25
Add diagnostics header

jerriep:
This PR adds the telemetry headers. 

For PCLs we have a problem retrieving information such as the .NET CLR version and OS name and version. 

PCLs also does not support retrieving the list of referenced DLLs, so as per Nathan's suggestion, I included the Nuget `packages.config` file as a resource and simply extract the referenced Nuget packages from there.

For the User-Agent I use the string `.NET/PCL`

Here is an example of the diagnostics header before it is Base64 encoded:

```
{  
   "dependencies":[  
      {  
         "name":"Microsoft.Bcl",
         "version":"1.1.10"
      },
      {  
         "name":"Microsoft.Bcl.Async",
         "version":"1.0.168"
      },
      {  
         "name":"Microsoft.Bcl.Build",
         "version":"1.0.21"
      },
      {  
         "name":"Microsoft.Net.Http",
         "version":"2.2.29"
      },
      {  
         "name":"Newtonsoft.Json",
         "version":"7.0.1"
      },
      {  
         "name":"PortableRest",
         "version":"3.1.0-Beta2a"
      }
   ],
   "environment":[  
      {  
         "name":".NET PCL",
         "version":"Profile111"
      }
   ],
   "name":"Auth0.ManagementApi.Client",
   "version":"1.0.0.0"
}
```

Let me know if any of these workarounds do not produce an acceptable result, so we can come up with alternatives.


src/Auth0.ManagementApi.Client/ApiConnection.cs
```diff
@@ -125,6 +128,13 @@ public ApiConnection(string token, string baseUrl)
                 }
             }
 
+            // Add the diagnostics header, unless user explicitly opted out of it
+            if (!object.ReferenceEquals(diagnostics, DiagnosticsHeader.Suppress))
+                request.AddHeader("Auth0-Client", diagnostics.ToString());
+
+            // Add the user agent
+            request.AddHeader("User-Agent", ".NET/PCL");
```
advancedrei: PortableRest automatically adds a UserAgent header. If you call SetUserAgent<ClientBase>() in your constructor, it will automatically add a User Agent header with both the current assembly version, and the version of PortableRest used. See https://github.com/advancedrei/PortableRest/blob/master/src/PortableRest/RestClient.cs#L133-L171

Also, there is a way to use Interfaces and a technique with PCLs called "bait and switch" to leverage NuGet to be able to swap out portable libraries with platform specific libraries. That technique could be used to get more specific diagnostic information. However, any diagnostics information sent down should be opt-in. I don't think just because I'm using your library means you have the right to scrape my system for information without my permission. Something to keep in mind.

jerriep: Yeah, I was thinking about the bait-and-switch technique too. We can look into it, but unfortunately I do not have access to the 2 Xamarin platforms, should be decide to go that route.

As far as the opt-in is concerned, I used the technique from the V1 library. Essenstially all you need to do is pass the `DiagnosticsHeader .Suppress` value to the constructor of `ManagementApiClient` and no diagnostics information will be sent at all.

Alternatively you can build up your own `DiagnosticsHeader` with whichever values you want, and I will send that along instead.

sandrinodimattia: Maybe we need to split the DiagnosticsHeader in 2 parts:

 - The opt in which contains system info like OS, .NET version, .... This is a nice to have for troubleshooting but not a necessity (it's more important on iOS)
 - One thing I think we would always need is to send the current version of the SDK. The reason is that if we make a breaking change (which we obviously avoid) or if we detect a security issue with an SDK, we need to be able to reach out to all customers using the impacted version of the SDK

PR #54, opened 2015-11-27
Add Auth API

jerriep:
Add the base, empty Auth API project, so @advancedrei and myself have the same base to work from.

@ntotten If you can fit this in quickly somewhere between your Thanksgiving celebrations it will be appreciated ;)


jerriep:
@ntotten The tests that failed is just about the Nuspecs which are not in this PR. They are still in another PR waiting to be accepted

PR #55, opened 2015-11-27
Organizational improvements + Normalized user profile

advancedrei:
- Added PortableRest to AuthAPI.
- Updated Bcl.Build to be consistent between projects.
- Moved DelegationRequest objects out of Core and into AuthAPI.
- Added missing User object properties from API responses.
- Alphabetized User object properties to make it easier to find "normalized" user profile properties that aren't documented as "normalized", such as last login date and IP.


advancedrei:
FYI these changes are in preparation for adding objects that can be posted from a currently-executing Rule to an external source. Those objects will utilize existing ones where possible. See https://github.com/auth0/rules/blob/master/rules/aspnet-webapi.md for context.

jerriep:
:+1:

PR #59, opened 2015-11-28
Objects supporting the Rules API

advancedrei:
This is to support a Rule snippet already in production. See
https://github.com/auth0/rules/blob/master/rules/aspnet-webapi.md


jerriep:
:+1:

advancedrei:
@ntotten Jerrie and I talked about this and he's ok with these changes. It may break the build, but go ahead and accept it anyways.

PR #60, opened 2015-12-01
Auth API reorganization

jerriep:
This PR merges all of Robert's changes into the existing work I have done so far. 

I have also reorganized to files and some of the projects to more easily share code between the Management and Authentication SDKs


jerriep:
@advancedrei Once @ntotten has accepted this PR you must please fetch from V2 branch of the origin and rebase your branch on top of that so you can have these changes.

If you make new changes and just carry on where your branch was last, we are going to run into issues.

PR #61, opened 2015-12-02
Work on Authentication API

jerriep:
* Add authorization method
* Add ability to launch the various Passwordless flows
* Obtain Access Token 
* ID Token delegation
* Get UserInfo
* Get TokenInfo
* All URL Builders (for methods that return 302s)
* SAML and WS-FED metadata


src/Auth0.AuthenticationApi.Client/Builders/WsFedUrlBuilder.cs
```diff
@@ -0,0 +1,34 @@
...
+        public WsFedUrlBuilder WithWhr(string connectionName)
+        {
+            AddQueryString("whr", $"urn:{connectionName}");
+
+            return this;
+        }
+
+        public WsFedUrlBuilder WithWtrealm(string clientId)
+        {
+            AddQueryString("wtrealm", $"urn:{clientId}");
```
sandrinodimattia: I would remove urn: from here. The problem is that a subscriber can override the wtrealm using client aliasses, allowing them to specify **http://foo** as the wtrealm instead of **urn:clientId**.

src/Auth0.AuthenticationApi.Client/Builders/WsFedUrlBuilder.cs
```diff
@@ -0,0 +1,34 @@
+﻿using JetBrains.Annotations;
+
+namespace Auth0.AuthenticationApi.Client.Builders
+{
+    public class WsFedUrlBuilder : UrlBuilderBase<WsFedUrlBuilder>
```
sandrinodimattia: I would add an additional method here `WithWctx` or `WithState` that sets the `wctx` to a value (this is the state).

Ideally this should be a string or a key value pair (that is then flattened to a string, like `xcrf=abc&ru=/foo` (make sure you also add url encoding here)

src/Auth0.AuthenticationApi.Client/Builders/SamlUrlBuilder.cs
```diff
@@ -0,0 +1,20 @@
+namespace Auth0.AuthenticationApi.Client.Builders
+{
+    public class SamlUrlBuilder : UrlBuilderBase<SamlUrlBuilder>
+    {
+        private string ConnectionName { get; set; }
+
+        public SamlUrlBuilder(string baseUrl, string client)
```
sandrinodimattia: I would add an additional method here `WithRelayState` or `WithState` that sets the `RleayState` to a value (this is the state).

Ideally this should be a string and an overload with a key value pair (that is then flattened to a string, like `xcrf=abc&ru=/foo` (make sure you also add url encoding here)

PR #64, opened 2015-12-11
Support "Delete a user's multifactor provider"

RaleS:


jerriep:
Thank you for the contribution @RaleS 

You code was based on a very old branch, so I just copied your code instead of trying to merge your PR. See e255e38

PR #66, opened 2015-12-17
Added device parameter to LoginUser request, since it's required

dgreenbean:


dgreenbean:
I would actually prefer to make device required, as it is required by the API, but that could break compilation for existing users.  This at least makes it possible to use the function.

jerriep:
Closing this PR. This is supported in the reworked SDK we did since V2

PR #68, opened 2015-12-18
Changing to reference client secret directly for delegation calls

dgreenbean:
The API does not require a client secret for all types of delegation.


jerriep:
Closing as no longer relevant, seeing this is based on old version of SDK.

PR #100, opened 2016-02-09
Fix Authentication Response NULL

MHamburg:
Could not convert JSON to object (properties were all null). This solved it for me as the returned json was different from the object format.

Let me know if this commit is correct, so I can use the nuget version.


jerriep:
Wow, thanks @MHamburg. I will updated the unit tests as well to ensure this is checked for

jerriep:
@MHamburg  New version available on Nuget:
https://www.nuget.org/packages/Auth0.AuthenticationApi/

Thanks for fix

MHamburg:
@jerriep yw, thx for the fast response!

PR #112, opened 2016-03-18
Missing properties for refreshtokens

MHamburg:
Please also allow the following properties to be implemented for the use of refreshtokens:

AuthenticationResponse:
- refresh_token

AccessToken
- expires_in


jerriep:
Thank you @MHamburg

MHamburg:
@jerriep, could you please update the nuget package? So we can use that version when deploying =)

jerriep:
@MHamburg V3.1.0 now released to Nuget :)

PR #119, opened 2016-06-14
Description on thread safety

nicosabena:
Add a description on thread safety for both clients.


jerriep:
@nicosabena Thank you :)

PR #126, opened 2016-10-03
Update UsernamePasswordLoginRequest.cs

FutureGUIs:
Add the missing device property in order to get a refresh token on login.


jerriep:
Thanks @FutureGUIs , I have pushed new version (3.4.1) to NuGet

FutureGUIs:
thanks a lot for the quick turnaround

Issue #147, opened 2017-05-19
ResourceServer association for Clients is incorrect


jerriep:
Reported by @rdowinton 

Submitting the following request:

```
ClientCreateRequest client_create = new ClientCreateRequest(); 
client_create.Name = Entity.FullName; 
client_create.ApplicationType = ClientApplicationType.NonInteractive; 
List<ResourceServer> ResourceServers = new List<ResourceServer>(); 
ResourceServer rs = new ResourceServer(); 
rs.Identifier = _auth0Conf.Identifier; 
rs.Id = _auth0Conf.IdIdentifier; 
rs.Scopes = new List<ResourceServerScope>(); 
ResourceServers.Add(rs); 
client_create.ResourceServers = ResourceServers.ToArray(); 
Jwt.SigningAlgorithm = "RS256"; 
Jwt.LifetimeInSeconds = 3600;

client_create.JwtConfiguration = Jwt;

var createTask = _auth0Api.Clients.CreateAsync(client_create); 
createTask.Wait();
```

Results in this exception:

```
Exception : Payload validation error: 'Additional properties not allowed: skip_consent_for_verifiable_first_party_clients,allow_offline_access,token_lifetime,signing_alg,id' on property resource_servers[0] ({description}).
```

The reason for this is that the association to ResourceServers in the `ClientBase` (from which `ClientCreateRequest` inherits) references the full schema for resource servers. This is not correct. If we look at the API schema for Clients on the Management API, the list of resource servers is a minumum set of attributes. The schema is definded as follows:

```
module.exports = {
  "type": "object",
  "properties": {
    "identifier": {
      "type": "string",
      "description": "The resource server identifier. It is recommended to use a URL for this",
      "minLength": 1
    },
    "scopes": {
      "type": "array",
      "items": {
        "type": "string",
        "minLength": 1,
        "uniqueItems": true,
        "description": "A scope to request from the user to access the resource server"
      },
      "description": "The scopes to request from the user to access the resource server"
    }
  },
  "required": ["identifier"],
  "additionalProperties": false
};
```

The Auth0.NET API should therefore be changed to create a new class (`ResourceServerAssociation` ?) which will only contain this minimum set of fields to prevent overposting and thus raising the error on the Management API.

Bad news: **This is a breaking change**
Good news: No-one is actually using this (other than the client who reported this to @rdowinton ), otherwise this would have been reported long time ago

PR #148, opened 2017-05-19
Add ClientResourceServerAssociation class

jerriep:
Add ClientResourceServerAssociation class for associating resource servers with clients

This is fix for #147

### DIFF
--- a/src/Auth0.ManagementApi/Models/ClientBase.cs
+++ b/src/Auth0.ManagementApi/Models/ClientBase.cs
@@ -119,9 +119,11 @@ namespace Auth0.ManagementApi.Models
         [JsonProperty("name")]
         public string Name { get; set; }
 
-
+        /// <summary>
+        /// 
+        /// </summary>
         [JsonProperty("resource_servers")]
-        public ResourceServer[] ResourceServers { get; set; }
+        public ClientResourceServerAssociation[] ResourceServers { get; set; }
 
         /// <summary>
         /// True to use Auth0 instead of the IdP to do Single Sign On, false otherwise.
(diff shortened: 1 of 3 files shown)


jerriep:
Seem we should use https://auth0.com/docs/api/management/v2#!/Client_Grants/post_client_grants instead. Will update

src/Auth0.ManagementApi/Models/ClientBase.cs
```diff
@@ -119,9 +119,11 @@ public abstract class ClientBase
         [JsonProperty("name")]
         public string Name { get; set; }
 
-
+        /// <summary>
+        ///
```
nicosabena: How about `A list of resource servers (APIs) that the client is authorized to request access tokens for, using the Client Credentials exchange.`?

src/Auth0.ManagementApi/Models/ClientResourceServerAssociation.cs
```diff
@@ -0,0 +1,23 @@
...
+    public class ClientResourceServerAssociation
+    {
+        /// <summary>
+        /// The resource server identifier.
+        /// </summary>
+        [JsonProperty("identifier")]
+        public string Identifier { get; set; }
+
+        /// <summary>
+        /// The scopes to request from the user to access the resource server.
```
nicosabena: I'd say it's more like "The scopes granted to the client to access the resource server."

src/Auth0.ManagementApi/Models/ClientResourceServerAssociation.cs
```diff
@@ -0,0 +1,23 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Auth0.ManagementApi.Models
+{
+    public class ClientResourceServerAssociation
+    {
+        /// <summary>
+        /// The resource server identifier.
```
nicosabena: I would add ` (API)` after `The resource server`, to match the dashboard and docs more closely.

PR #150, opened 2017-06-09
 Add refresh token property in user identity

jmangelo:
This will allow refresh tokens to be obtained when retrieving users from the Management API if the appropriate scopes are specified in the access token.


jerriep:
@jmangelo Please find V4.3.0 available on GitHub. Ads both "RefreshToken` and `ExpiresIn` properties to `Identity`

PR #155, opened 2017-10-13
Add missing properties to LogEntry

barclayadam:
Adds properties present in API return but missing from C# model


jerriep:
@barclayadam Thank you, I am following up internally on these fields since they do not appear in the API docs

barclayadam:
@jerriep Any update on this? We use the information present here in a production app and don't want to be falling behind SDK master.

jerriep:
@barclayadam Sadly not. I have asked clarification from the team responsible for that a couple of times before I left Auth0's permanent employ end of October and did not get a reply.

I have also subsequently in my capacity as freelancer enquired about this a few times, and still no answer. Unfortunately, in my current capacity, I do not have direct access to that team anymore. I keep prompting through the channels I have at my disposal every few months, but no luck.

If you are a paying Auth0 customer, I suggest that you also raise a formal support request for this through the ticketing system, and reference this GH issue. Maybe that way we can get someone to respond.

jerriep:
@barclayadam Have you had a chance to think about this?

jerriep:
@barclayadam Pinging you again about this....

barclayadam:
Hi @jerriep sorry for the radio silence. 

I can see your suggestion working, although I believe as these are well-defined properties having them available as strongly typed makes more sense. These are not ad-hoc add-anything properties like your blog example.

If they do not exist for a log entry they would be null. That can be checked for in the same way as AdditionalData.ContainsKey or LogType == "whatever", but we maintain the actual types instead of having to further parse `JToken`s.

We have come across so much missing in the .NET SDK compared to the docs. From tenant settings to resourceserver/api properties missing I wonder why the SDK is not auto-generated in some way.

barclayadam:
In addition, leaving them as just AdditionalData requires more investigation by end-users to determine what is available and what can be done with the tokens. This part of the API seems generally under-developed from documentation and SDK POV

jerriep:
> as these are well-defined properties

The problem is exactly that they are **not** well-defined. They are not listed in the API docs, and I could not get clarification from the responsible Auth0 team.

barclayadam:
well-defined was probably the wrong terminology. Auth0-defined is more what I was looking for. These are well-known within the Auth0 team, as opposed to end-user controlled properties that may be attached to user metadata/profile.

I see it as either the docs & SDK gets updated and these properties documented, or they get removed from the response. This middle-ground means I cannot rely on this at all, making my app liable to sudden breakage.

jerriep:
I can ask again for confirmation of those properties. But seeing as I did not get any last time, I am not positive that this time around will give a better result :(

barclayadam:
Unfortunately, I agree. Yesterday I raised a support issue directly in an attempt to get some answers as well

saltukalakus:
Hi @jerriep in case, you didn't receive, I send you a message from a ticket. Feel free to reply from there or this thread for me chase the questions internally.

jerriep:
@saltukalakus This PR is adding support for fields in the payloads received from the `/api/v2/logs` endpoints. For example, here is a payload when I call that endpoint on my own Auth0 tenant:

```json
[
  {
    "date": "2018-07-30T02:31:07.594Z",
    "type": "seccft",
    "description": "",
    "connection_id": "",
    "client_id": "11dV61U...",
    "client_name": "API Explorer Application",
    "ip": "35.167...",
    "hostname": "jerrie.auth0.com",
    "user_id": "",
    "user_name": "",
    "audience": "https://jerrie.auth0.com/api/v2/",
    "scope": null,
    "_id": "90020180730023107595263179019441474202980571687761739874",
    "log_id": "90020180730023107595263179019441474202980571687761739874",
    "isMobile": false,
    "user_agent": "Other 0.0.0 / Other 0.0.0"
  }
]
```

However, the docs for that endpoint state that this is a sample response:

```json
[
  {
    "date": "2016-02-23T19:57:29.532Z",
    "type": "sapi",
    "client_id": "AaiyAPdpYdesoKnqjj8HJqRn4T5titww",
    "client_name": "My application Name",
    "ip": "190.257.209.19",
    "location_info": {},
    "details": {},
    "user_id": "auth0|56c75c4e42b6359e98374bc2"
  }
]
```

As you can see, there are various fields returned in the actual response which are not in the sample, documented response. For example `connection_id`, `user_name`, `user_agent` etc.

This PR was created to add those fields to the Auth0 .NET SDK. However, I cannot accept this unless those fields are officially documented as valid fields being returned in the payload. I have made this mistake in the past and it created problems since in that particular instance the fields were later on removed and when I questioned it, I was told I should have never added it because it was not documented. It was just an experiment.

So my criteria for adding fields is that it **must be documented**.

Now, before I left Auth0 as a permanent employee in October last year, I tried on various occasions to get clarification on this with the relevant crew (Crew 2, I believe). I never got an answer from them. I also subsequently as contracted tried to get a response, but I was not successful.

So, the bottom line is that we need confirmation from the crew responsible for the Management API as to what the official, documented list of fields are which are returned by that API endpoint.

jerriep:
I received confirmation that the list of fields are as follows:

```
[ 'type',
 'date',
 'client_id',
 'client_name',
 'user_id',
 'description',
 'user_name',
 'details',
 'connection',
 'user_agent',
 'hostname',
 'impersonator_user_id',
 'impersonator_user_name',
 'strategy',
 'strategy_type',
 'auth0_client',
 'connection_id',
 'isMobile',
 'ip',
 '_id' ]
```

jerriep:
@barclayadam I pushed v5.9.0 to NuGet with these changes. Apologies again for the tremendous amount of time this took.

Issue #165, opened 2017-11-30
ApiConnection is not disposing of HttpClient properly, while requiring a dynamic token value in constructor


jaccus:
Either let configure token dynamically, on method call (then can make the ApiConnection a singleton and keep it global), or properly dispose of it (since need to recreate clients as often as tokens expire, e.g., every few hours).

jerriep:
@jaccus I assume you are referring to `ManagementApiClient`?

I can look into adding a `SetToken` method which will update the token on the underlying `ApiConnection`. Will that work for you?

jaccus:
This goes back to me having to use Guardian features from the SDK. While the `Guardian` property was hidden on `ManagementApiClient` I started using `GuardianClient` directly, which in turn will cause memory leaks due to `ApiConnection` not disposing of `HttpClient`'s properly.

Since `ManagementApiClient` is using clients under the hood anyways, guess this issue does bubble up to `ManagementApiClient`.

jerriep:
*Please* do not use `GuardianClient`. Your code will break when I update that API. 

Also, `HttpClient` should not be disposed. See https://stackoverflow.com/questions/15705092/do-httpclient-and-httpclienthandler-have-to-be-disposed and https://aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/

jaccus:
Yes it should not, but your current implementation (where you instantiate HttpClient inside of your ApiConnection) makes it necessary to update the token value.

jaccus:
Another thing is, if you encapsulate an instantiation of an IDisposable object in your constructor, it's a good practice to implement IDisposable yourself (in this case IApiConnection), where in the Dispose method you dispose of HttpClient.

yngndrw-sage:
The documentation page for this SDK suggests that the clients are not threadsafe and therefore shouldn't be shared:
http://auth0.github.io/auth0.net/#client-reuse-and-thread-safety

The issue with this is that there's no way to either clean-up or re-use the underlying ```HttpClient```. Maybe ```ApiConnection``` could be passed in to the API clients and token changing allowed ?

jerriep:
@yngndrw-sage 

That is specifically related to when using the `GetLastApiInfo()` method which retrieves the status for the last API call. That particular "feature" has nothing to do with how we use or dispose of the `HttpClient` internally.

If you do not use that, then you can certainly use it in multi-threaded environments without issue.

yngndrw-sage:
[...]

PR #166, opened 2017-11-30
restore commented out Guardian client to ManagementApiClient

### LINKED ISSUE #164: Guardian property is hidden on ManagementApiClient even though, GuardianClient is available
I would like to use Guardian client, but I don't want to deal with IApiConnection and stuff.

Please uncomment your code that was poorly "hidden" in [https://github.com/auth0/auth0.net/commit/faa7af5228ee3415a2e5381d220041a0513a21db](https://github.com/auth0/auth0.net/commit/faa7af5228ee3415a2e5381d220041a0513a21db) :)

Would also appreciate some info on why was this hidden? Is this a buggy functionality? I'd rather not hit the Auth0 APIs bypassing this .NET library to keep it future-proof.

jaccus:
Resolves #164

### DIFF
--- a/src/Auth0.ManagementApi/IManagementApiClient.cs
+++ b/src/Auth0.ManagementApi/IManagementApiClient.cs
@@ -42,7 +42,7 @@ namespace Auth0.ManagementApi
         /// <summary>
         /// Contains all the methods to call the /guardian endpoints.
         /// </summary>
-        //IGuardianClient Guardian { get; }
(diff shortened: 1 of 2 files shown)


jerriep:
As per comment on #165 , this is not something to we can do now, and not as simple as just exposing that property. A full implementation of the Guardian endpoints need to be done.

I am therefore closing this PR

PR #170, opened 2018-02-12
Add multi-target support for net471. Fixes #169

### LINKED ISSUE #169: Why do the NuGet packages depend on System.Net.Http and System.Net.Primitives
In the nuspec files, the dependencies System.Net.Http and System.Net.Primitives are listed, but the projects itself don't need that dependency.

This gives all kind of problems in net462 / net471 scenario's see my [post on StackOverflow](https://stackoverflow.com/questions/48060230/system-reference-troubles-when-introducing-netstandard-library-dependency) which could be avoided if the Auth0 packages didn't have that dependency, or at least add an exception for `net471` that doesn't need those dependencies.

Is it possible to add support for net471, without having to add additional NETstandard packages?

huysentruitw:


jerriep:
@huysentruitw I am just going to hold off on this one, since there is another change coming which removes the nuspec files completely:
https://github.com/jerriep/auth0.net/tree/ci

I can look into addressing your issue in that one

Can you use the workaround in the StackOverflow answer for the time being...?

huysentruitw:
We had a lot of dependency conflicts (not only System.Net.Http) which I wanted to resolve in a clean way (and that's not adding those packages everywhere IMO).

The only clean way to get rid of this problem is that every library-owner deals with net471 separately.

Currently we are building these Auth0 packages with this fix and publish them to an internal NuGet feed. It works like a charm.

So, even if you decide to remove the nuspec files, you'd still need the csproj changes from this PR. Another solution could probably be to target netstandard 2.0 instead of 1.1 but can cause issues for other customers.

jerriep:
Adding netstandard 2.0 in addition to  netstandard 1.1 may be a good solution. I will look into this later in the week.

huysentruitw:
Thanks for considering this @jerriep 👍

PR #175, opened 2018-03-13
Add RefreshTokenRequest Audience to actual request

plucth:
I have a rule that need that info for scope attribution.
And the fact that the attribute exist in the model and is not used is kind of confusing.

### DIFF
--- a/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
+++ b/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
@@ -335,9 +335,13 @@ namespace Auth0.AuthenticationApi
                 { "grant_type", "refresh_token" },
                 { "refresh_token", request.RefreshToken },
                 { "client_id", request.ClientId },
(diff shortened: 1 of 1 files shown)


plucth:
I'm talking about line 14 in src/Auth0.AuthenticationApi/Models/RefreshTokenRequest.cs

And the parameter is in the "OIDC-conformant (token endpoint)" exemple of https://auth0.com/docs/api-auth/tutorials/adoption/refresh-tokens

jerriep:
Aah yes, I understand. I am however not sure (1) why that property was added to `RefreshTokenRequest` in the first place, and (2) whether it is indeed valid passing an `audience` parameter to the `oauth/token` endpoint when requesting a refresh token.

Let me ask my contact at Auth0 for clarification

plucth:
No problem, thank you for your time

jerriep:
@plucth The will be added to the upcoming version 5 of the library. Will be in the next week or so hopefully. I am just busy with more cleanup for that release.

PR #190, opened 2018-06-29
Inject HttpClient into AuthenticationApiClient 

davidallyoung:
In reference to #165 I've spiked out opening up the AuthenticationApiClient to allow consumers to inject their own HttpClient that they can manage the lifetime of per their application requirements.

This is to help comply with what is given on MSDN regarding HttpClient socket exhaustion: https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=netframework-4.7 `HttpClient is intended to be instantiated once and re-used throughout the life of an application. Instantiating an HttpClient class for every request will exhaust the number of sockets available under heavy loads. This will result in SocketException errors.`

Let me know if you'd like for me to extend this into the ManagementApiClient before acceptance or if you'd like specific test cases written.

### DIFF
--- a/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
+++ b/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
@@ -25,12 +25,6 @@ namespace Auth0.AuthenticationApi
         {
             _baseUri = baseUri;
 
-            // If no diagnostics header structure was specified, then revert to the default one
-            if (diagnostics == null)
-            {
-                diagnostics = DiagnosticsHeader.Default;
-            }
-
             Connection = connection ?? throw new ArgumentNullException(nameof(ApiConnection));
         }
 
@@ -41,7 +35,7 @@ namespace Auth0.AuthenticationApi
         /// <param name="diagnostics">The diagnostics.</param>
         /// <param name="handler">The <see cref="HttpMessageHandler"/> which is used for HTTP requests</param>
         public AuthenticationApiClient(Uri baseUri, DiagnosticsHeader diagnostics, HttpMessageHandler handler)
-            : this(baseUri, diagnostics, new ApiConnection(null, baseUri.AbsoluteUri, diagnostics, handler))
+            : this(baseUri, diagnostics, new ApiConnection(null, baseUri.AbsoluteUri, diagnostics ?? DiagnosticsHeader.Default, handler))
         {
         }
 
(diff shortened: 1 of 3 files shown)


jerriep:
@lucascrespo I'll look into this in detail next week. On the surface it seems ok, so you can add for Management API as well if you'd like.

I need to get my mind dedicated on this though so I can understand all possible implications. Will make time for it next week.

davidallyoung:
@jerriep Awesome, thanks so much!

jerriep:
@nicosabena When you have time, can you please look at this. The idea is that the lifetime of `HttpClient` can be managed outside of the `AuthenticationApiClient` and `ManagementApiClient` instances. This will also allow better integration with the `HttpClientFactory` (https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-2.1) in ASP.NET Core

nicosabena:
Looks good to me, at least for a minor version release. On the next major release we should make these classes `IDisposable` and implement the [Dispose pattern](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/dispose-pattern) if we will keep creating an instance of `HttpClient` when not provided (which we probably should, for usability).
Thanks @davidallyoung! And I think we can gratefully accept your offer of extending this to `ManagementApiClient`.

davidallyoung:
Hey guys, I've updated this for the ManagementApiClient to allow injecting an HttpClient as well. Sorry for it taking a few, had a couple of 🔥 to put out and 4th of July in the U.S. Thanks for letting me contribute!

My two cents on the IDisposable implementation is that I would make the ApiConnection be the item that is disposable and re-work these constructors to take in ApiConnections, or no longer automatically create HttpClients for consumers. This could let someone register an ApiConnection in their DI framework, poor man's DI, or however they want, and dump it off when they know they don't need that HttpClient anymore. (I think I like the next option more though)

I think the risk in making it IDisposable (and allow injection of HttpClient) is that it could confuse someone that isn't careful in what they're doing, IE they inject their singleton HttpClient into an Auth0 ApiClient that's in a tight using statement and now they've broken their app when the Auth0 client gets dumped off :(.  Perhaps the API clients should not implement IDisposable at all and instead force consumers to give an HttpClient so that the dependency and responsibility is clear.

jerriep:
Thank you @davidallyoung . I will take care of this next week.

jerriep:
Thanks again for the contribution @davidallyoung. I hope to release a new version soon.

jerriep:
Version 5.7.0 is now published with these changes

PR #191, opened 2018-07-02
Implement paging phase 2

jerriep:
This PR implements Phase 2 of adding pagination to Management API endpoints. Done in this PR are:

[x] Client Grants
[ ] User Grants
[x] Resource Servers
[x] Rules

`IClientGrantsClient` and `IRulesClient` has introduced breaking changes because previously all parameters were also optional. So not passing any parameters will throw

```
[CS0121] The call is ambiguous between the following methods or properties: 'IClientGrantsClient.GetAllAsync(string, string)' and 'IClientGrantsClient.GetAllAsync(int?, int?, bool?, string, string)'
```

Need to discuss how to handle this. One way is to simply append paging parameters to previous `GetAllAsync` method and change the return type of that method to `Task<IPagedList<T>>`. Since `IPagedList<T>` inherits from `IList<T>` this should not break any code. The only problem with this is that we will be inconsistent with other paged methods where the paging parameters occur as the first set of parameters

**Edit**: Also see extra comments below for additional thoughts

### DIFF
--- a/.idea/.idea.Auth0.Net/.idea/contentModel.xml
+++ b/.idea/.idea.Auth0.Net/.idea/contentModel.xml
@@ -191,6 +191,8 @@
             <e p="EmailTemplateUpdateRequest.cs" t="Include" />
             <e p="EmailVerificationTicketRequest.cs" t="Include" />
             <e p="EncryptionKey.cs" t="Include" />
+            <e p="GetClientGrantsRequest.cs" t="Include" />
+            <e p="GetRulesRequest.cs" t="Include" />
             <e p="GuardianEnrollment.cs" t="Include" />
             <e p="GuardianEnrollmentStatus.cs" t="Include" />
             <e p="GuardianFactor.cs" t="Include" />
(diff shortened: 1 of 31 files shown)


jerriep:
One other thing we can do as well is to create a request class for all of the `GetAllAsync` methods in the library. Some of the existing methods already have a long list of parameters. Case in point:
https://github.com/auth0/auth0.net/blob/236a7e4a9bbda8fdc2aa92af71d23a7e617e32e1/src/Auth0.ManagementApi/Clients/IUsersClient.cs#L62-L63

```csharp
Task<IPagedList<User>> GetAllAsync(int? page = null, int? perPage = null, bool? includeTotals = null, string sort = null, string connection = null, string fields = null,
bool? includeFields = null, string q = null, string searchEngine = null);
```

So we can instead replace it with something like

```csharp
Task<IPagedList<User>> GetAllAsync(GetUsersRequest request);
```

where `GetUsersRequest` is a class containing all the current parameters as properties. This way we can introduce new method signatures for all the existing `GetAllAsync` methods without introducing breaking changes. We can also mark the existing methods as `[Obsolete]`, giving customers time to migrate to the new request object structure.

PR #195, opened 2018-07-12
Possibility to add an auth0-forwarded-for header to ROPC token requests

ddeschepper:
The current option to add a auth0-forwarded-for header to ROPC token requests, is to implement a custom Http​Message​Handler and overriding SendAsync. However, this removes the possibility to provide a reusable HttpClient. Since this header is an official feature of ROPC token requests, I think it is ok to add the option of using it to the ResourceOwnerTokenRequest.

I looked for any tests that should be adapted for this change, but could not find any. Please point me in the right direction if any more work needs to be done.

### DIFF
--- a/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
+++ b/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
@@ -372,7 +372,14 @@ namespace Auth0.AuthenticationApi
                 parameters.Add("realm", request.Realm);
             }
 
-            return Connection.PostAsync<AccessTokenResponse>("oauth/token", null, parameters, null, null, null, null);
+            var headers = new Dictionary<string, object>();
+
+            if (!string.IsNullOrEmpty(request.ForwardedForIp))
+            {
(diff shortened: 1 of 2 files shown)


jerriep:
@ddeschepper Can you please provide me with more information regarding this? I do not quite understand the use and context of this header.

Also, what does the ROPC acronym stand for and what is the "official feature of ROPC token requests" you are referring to?

jerriep:
Thank you for the clarification @ddeschepper . I just wanted to be sure that `ResourceOwnerTokenRequest` is the correct place to put this and that it should not be available to a wider range of requests. It seems per the document you linked, as well as the [Management API Explorer](https://auth0.com/docs/api/authentication#resource-owner-password), that this is the only place it is relevant.

Thank you for this contribution. I hope to make it available in a release next week some time.

jerriep:
@ddeschepper Version 5.8.0 has been pushed to NuGet and will be available for you to install shortly

ddeschepper:
@jerriep thanks :)

Issue #200, opened 2018-07-25
Expose server time from headers through GetLastApiInfo


Vaskinn:
We're having a service that we want to automatically slow down its requests if we reach auth0 rate limits. Unfortunately we cannot make use of `GetLastApiInfo().RateLimit.Reset` because server time from Date header in the reponse isn't easily available to compare it to.

I suggest adding a date property to `ApiInfo` and setting its value in `ApiInfoParser.Parse()`

If you agree this is a useful idea I can create a pull request that does this.

jerriep:
The time given back in `GetLastApiInfo().RateLimit.Reset` it the UTC time. So you can just convert your own server's time to UTC and then figure out the difference.

Will that now work for you?

Vaskinn:
Unfortunately using the time from `Reset` alone needs the time on our server to be perfectly in sync with the time on the auth0 servers. I don't think I can trust them to be. It feels a lot safer to use the difference between two time values that are both provided by the same server.

PR #201, opened 2018-07-25
Add ServerTime to ApiInfo

Vaskinn:
This adds ServerTime to ApiInfo as discussed in #200.

Had to use a nullable DateTimeoffset because the header could technically have been left out by the server.


jerriep:
Thank you @Vaskinn 

I will try and do a new release to NuGet later today

cyberezra:
Pill request

jerriep:
@Vaskinn Version 5.8.0 has been pushed to NuGet with this fix and will be available for you to install shortly

PR #208, opened 2018-09-10
Add token_lifetime_for_web to ResourceServerBase

barclayadam:
Adds a missing property from /resource-servers API endpoint


dmarjoram:
Also needed this, thanks

dmarjoram:
An word on an updated build / nuget package? Thanks

luisrudge:
@dmarjoram It's coming! We're changing the release process, but I think we'll have something out by the end of this week. Sorry to keep you waiting.

damieng:
Shipped in 5.11.0 (out now)

PR #210, opened 2018-09-19
Add PhoneVerified property to model

adjkilgore:
This will allow the Auth0.Net SDK to support creating Users on an SMS connection, without them receiving SMS messages - as per the Management API v2 docs.


damieng:
This PR seems to be doing a lot more than the subject says.  We like a PR to be highly focused so I'm going to close this PR for now unless you want to create one that is just adding the PhoneVerified to the model.

PR #240, opened 2019-03-20
Update management API to include GuardianClient

### LINKED ISSUE #231: Missing mark_email_as_verified field in PasswordChangeTicketRequest
According to API docs, there should be a "mark_email_as_verified" boolean field for PasswordChangeTicketRequest. It is missing for now.

https://auth0.com/docs/api/management/v2#!/Tickets/post_password_change

damieng:
There are a bunch of changes to the management API that have not been captured in the Auth.Net SDK - gathering them here.

Fixes #231 and #164


src/Auth0.ManagementApi/Clients/IGuardianClient.cs
```diff
@@ -1,7 +1,8 @@
-﻿using System.Collections.Generic;
+﻿using Auth0.ManagementApi.Models;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 
-namespace Auth0.ManagementApi.Models
```
lbalmaceda: Is this public? if so, isn't it a breaking change?

src/Auth0.ManagementApi/Models/CreateGuardianEnrollmentTicketRequest.cs
```diff
@@ -5,20 +5,20 @@ namespace Auth0.ManagementApi.Models
...
         [JsonProperty("email")]
         public string EmailAddress { get; set; }
 
         /// <summary>
-        ///     Indicates whether an email must be sent to the user to start the enrollment.
+        /// Indicates whether an email must be sent to the user to start the enrollment.
         /// </summary>
         [JsonProperty("send_mail")]
-        public bool MustSendMail { get; set; }
+        public bool? MustSendMail { get; set; }
```
lbalmaceda: this change is making it null safe? 🤔

tests/Auth0.ManagementApi.IntegrationTests/TicketsTests.cs
```diff
@@ -62,7 +62,7 @@ public async Task Test_tickets_sequence()
             {
                 UserId = _user.UserId,
                 ResultUrl = "http://www.nonexistingdomain.aaa/success",
-                NewPassword = Password
+                MarkEmailAsVerified = true
```
lbalmaceda: Would this line fail if the method `MarkEmailAsVerified` does not exists? I think you also added `IncludeEmailInRedirect` and would be nice to include it on the tests.

src/Auth0.ManagementApi/Clients/IGuardianClient.cs
damieng: It would be however IGuardianClient and GuardianClient were never officially added to the main class that people use to access this functionality.  People would have had to have subclassed ours and added it in themselves.  If they went through that then they'll get a compile error but I doubt anyone ever did.

src/Auth0.ManagementApi/Models/CreateGuardianEnrollmentTicketRequest.cs
damieng: It's actually the opposite.  This can now be null, true or false.  This is required whenever we have an API that has optional fields on the REST API otherwise there is no way of not sending the value.

tests/Auth0.ManagementApi.IntegrationTests/TicketsTests.cs
damieng: I'll add IncludeEmailInRedirect to this test.  The app would fail to compile if MarkEmailAsVerified was missing.

src/Auth0.ManagementApi/Clients/IGuardianClient.cs
lbalmaceda: I'd argue that. If the class was already pushed and scoped as public, people could have used it as you mention. e.g. The user that reported this initially. For them this is going to be a breaking change which, if we stick to semver, we must release a new major. If you can avoid this change that would be awesome. I'll leave the call to you.

damieng: Let's bump the major then.

src/Auth0.ManagementApi/Models/CreateGuardianEnrollmentTicketRequest.cs
damieng: Oh, and making something nullable is non-breaking because of the way the compiler/language works :)

Issue #246, opened 2019-03-31
Permissions endpoints


christophano:
I see that role endpoints were recently included (#241), but this change didn't include the permissions endpoints that exist on roles and users. Are there any plans to include them?

damieng:
Merged in for 6.0 - thanks for spotting this.

christophano:
Fantastic stuff! I've been working off a fork where I implemented this (on roles, at least) to get me going, but I'm pleased that official support is on its way!

PR #248, opened 2019-04-01
Add role permission endpoints to RoleClient

damieng:
Adds the permission endpoints as requested in #246 

Also fixes a couple of XML doc comments and makes ResourceServer.TokenLifetime nullable so you can patch it without changing it.

### DIFF
--- a/src/Auth0.ManagementApi/Clients/ClientBase.cs
+++ b/src/Auth0.ManagementApi/Clients/ClientBase.cs
@@ -8,7 +8,7 @@ namespace Auth0.ManagementApi.Clients
     public class ClientBase
     {
         /// <summary>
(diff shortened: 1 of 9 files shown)


src/Auth0.ManagementApi/Clients/ClientBase.cs
```diff
@@ -8,7 +8,7 @@ namespace Auth0.ManagementApi.Clients
     public class ClientBase
     {
         /// <summary>
-        /// The <see cref="IApiConnection"/> which is used to make all REST calls.
+        /// The <see cref="IApiConnection"/> which is used to make all HTTP API calls.
```
damieng: Change as requested by Josh in previous PR but not directly related to this.

PR #255, opened 2019-05-06
Add missing user permissions endpoints

damieng:
1. Adds `GET users/{id}/permissions` endpoint as `Users.GetPermissionsAsync(...)`
2. Adds `POST users/{id}/permissions` endpoint as `Users.AssignPermissionsAsync(...)`
3. Adds `DELETE users/{id}/permissions` endpoint as `Users.RemovePermissionsAsync(...)`

Also adjusts the existing naming on RolesClient permissions endpoints to be consistent with the API documentation.  Associate and Unassociate is now Assign and Remove.

### DIFF
--- a/src/Auth0.ManagementApi/Clients/UsersClient.cs
+++ b/src/Auth0.ManagementApi/Clients/UsersClient.cs
@@ -318,7 +318,7 @@ namespace Auth0.ManagementApi.Clients
         }
 
         /// <summary>
-        /// Removes permissions assign to a user.
+        /// Removes permissions assigned to a user.
         /// </summary>
         /// <param name="id">The ID of the user to remove permissions from.</param>
         /// <param name="request">A <see cref="AssignPermissionsRequest" /> containing the permission identifiers to remove from the user.</param>
--- a/src/Auth0.ManagementApi/Clients/UsersClient.cs
+++ b/src/Auth0.ManagementApi/Clients/UsersClient.cs
@@ -286,5 +286,47 @@ namespace Auth0.ManagementApi.Clients
                 {"id", id}
             });
         }
+
+        /// <summary>
+        /// Get the permissions assigned to the user.
+        /// </summary>
+        /// <param name="id">The id of the user to obtain the permissions for.</param>
+        /// <param name="pagination">Specifies <see cref="PaginationInfo"/> to use in requesting paged results.</param>
+        /// <returns>An <see cref="IPagedList{Permission}"/> containing the assigned permissions for this user.</returns>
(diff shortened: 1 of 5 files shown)


src/Auth0.ManagementApi/Clients/UsersClient.cs
```diff
@@ -286,5 +286,47 @@ public Task<User> UpdateAsync(string id, UserUpdateRequest request)
                 {"id", id}
             });
         }
+
+        /// <summary>
+        /// Get the permissions assigned to the user.
+        /// </summary>
+        /// <param name="id">The id of the user to obtain the permissions for.</param>
```
joshcanhelp: "id" => "ID"

src/Auth0.ManagementApi/Clients/UsersClient.cs
```diff
@@ -286,5 +286,47 @@ public Task<User> UpdateAsync(string id, UserUpdateRequest request)
...
+        /// <param name="request">A <see cref="AssignPermissionsRequest" /> containing the permission identifiers to assign to the user.</param>
+        /// <returns>A <see cref="Task"/> that represents the asynchronous assignment operation.</returns>
+        public Task AssignPermissionsAsync(string id, AssignPermissionsRequest request)
+        {
+            return Connection.PostAsync<object>("users/{id}/permissions", request, null, null,
+                new Dictionary<string, string> { { "id", id }, }, null, null);
+        }
+
+        /// <summary>
+        /// Removes permissions assign to a user.
```
joshcanhelp: "assign" => "assigned"

src/Auth0.ManagementApi/Clients/UsersClient.cs
damieng: There is a mix of both throughout the codebase.  Probably deserves a separate clean-up PR.

damieng:
> I'm wondering ... this endpoint is a little confusing (and our documentation definitely needs improvement) in that would have send an object, it has to have a `permissions` property, that property must be an array, that array must contain objects, and those objects must each have a property of `set to the permission and` set to the API identifier. Could we validate what they use in `request` to make sure we're sending it properly? 

It's a strongly typed language so you can't physically send anything malformed other than an empty or nulled array.  Also performing extra checks would be inconsistent with the rest of the SDK and provide more difficulty going forward when we switch to code-generated.

joshcanhelp:
@damieng - But couldn't you at least receive an array in the method and hand that to a `permissions` property that you send?

damieng:
Would be inconsistent with all other APIs and cause breaking changes if/when new properties are added to the request.

PR #256, opened 2019-05-07
Prepare 6.1.0

damieng:
- [x] Version bump assemblies
- [x] Rebuild documentation

Review notes: Review by commit, ignore regeneration of API docs.

### DIFF
--- a/src/Auth0.ManagementApi/Clients/RolesClient.cs
+++ b/src/Auth0.ManagementApi/Clients/RolesClient.cs
@@ -178,13 +178,6 @@ namespace Auth0.ManagementApi.Clients
                 }, null, null, new PagedListConverter<Permission>("users"));
         }
 
-        // TODO: Remove in 7.0.0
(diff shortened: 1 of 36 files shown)


src/Auth0.ManagementApi/Clients/RolesClient.cs
```diff
@@ -178,13 +178,6 @@ public Task<IPagedList<Permission>> GetPermissionsAsync(string id, PaginationInf
                 }, null, null, new PagedListConverter<Permission>("users"));
         }
 
-        // TODO: Remove in 7.0.0
-        [Obsolete("Use AssignPermissionsAsync instead")]
```
lbalmaceda: Is noted as "remove in 7.0.0" but this is 6.1.0

damieng:
Yes, it turns out I never shipped those methods - they missed 6.0.0 so no
need to deprecate them.

PR #268, opened 2019-06-12
Add missing Tenant settings

damieng:
Adds various settings to the Tenant APIs that were missing from the .NET SDK.

### DIFF
--- a/src/Auth0.ManagementApi/Models/TenantChangePassword.cs
+++ b/src/Auth0.ManagementApi/Models/TenantChangePassword.cs
@@ -1,0 +1,22 @@
+﻿using Newtonsoft.Json;
+
+namespace Auth0.ManagementApi.Models
+{
+    /// <summary>
+    /// Change Password page customization.
+    /// </summary>
+    public class TenantChangePassword
+    {
+        /// <summary>
+        /// true to use the custom change password html, false otherwise (default: false).
+        /// </summary>
+        [JsonProperty("enabled")]
+        public bool Enabled { get; set; }
+
+        /// <summary>
+        /// Replace default change password page with a custom HTML (Liquid syntax is supported).
+        /// </summary>
+        [JsonProperty("html")]
+        public string Html { get; set; }
+    }
+}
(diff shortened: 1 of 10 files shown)


src/Auth0.ManagementApi/Models/TenantGuardianMfa.cs
```diff
@@ -0,0 +1,22 @@
+﻿using Newtonsoft.Json;
+
+namespace Auth0.ManagementApi.Models
+{
+    /// <summary>
+    /// Guardian page customization.
+    /// </summary>
+    public class TenantGuardianMfa
```
joshcanhelp: Suggest `TenantGuardianMfaPage` for this (like `TenantErrorPage` above)

src/Auth0.ManagementApi/Models/TenantChangePassword.cs
```diff
@@ -0,0 +1,22 @@
...
+    public class TenantChangePassword
+    {
+        /// <summary>
+        /// true to use the custom change password html, false otherwise (default: false).
+        /// </summary>
+        [JsonProperty("enabled")]
+        public bool Enabled { get; set; }
+
+        /// <summary>
+        /// Replace default change password page with a custom HTML (Liquid syntax is supported).
```
joshcanhelp: Remove "a"

src/Auth0.ManagementApi/Models/TenantGuardianMfa.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        /// <summary>
+        /// true to use the custom html for Guardian page, false otherwise (default: false).
+        /// </summary>
+        [JsonProperty("enabled")]
+        public bool Enabled { get; set; }
+
+        /// <summary>
+        /// Replace default Guardian page with a custom HTML (Liquid syntax is supported).
```
joshcanhelp: Remove "a"

src/Auth0.ManagementApi/Models/TenantGuardianMfa.cs
damieng: Will do.

src/Auth0.ManagementApi/Models/TenantChangePassword.cs
damieng: Actually like that in the api-docs :(  Will correct upstream too.

PR #269, opened 2019-07-04
Add snyk policy file

lbalmaceda:
### Changes

Add a snyk policy file to ignore vulnerability reports that don't affect this SDK


lbalmaceda:
@damieng let's see if this fixes the reports

PR #272, opened 2019-07-19
Correct numeric type for lifetimes on TenantSettings

### LINKED ISSUE #271: Typing mismatch in TenantSettingsBase IdleSessionLifetime and SessionLifetime properties
### Description

The TenantSettingsBase class stores IdleSessionLifetime and Sessionlifetime as nullable ints -- and these values are set in minutes in the management site. When calling TenantSettings.GetAsync(), the values are returned as hours, and as a result I am receiving 0.25 for IdleSessionLifetime when the IdleSessionLifetime is set to 15 minutes. Naturally attempting to deserialize a floating-point number into an int throws an exception:

`Unhandled Exception: Newtonsoft.Json.JsonReaderException: Input string '0.25' is not a valid integer. Path 'idle_session_lifetime', line 1, position 562.`

If the intended behavior is indeed for the Auth0 management APIs to represent the session lifetime fields in hours, then the TenantSettingsBase class should be updated to store these fields as decimals rather than ints. 

### Environment

Please provide the following:

- Version of Auth0.NET used: 6.3.0
- Version of .NET used: .NET Standard 2.0

### Reproduction
Query the TenantSettings of a tenant with an SessionLifetime/IdleSessionLifetime not divisible by 60.
[stack_trace.txt](https://github.com/auth0/auth0.net/files/3408634/stack_trace.txt)

damieng:
This corrects the type of SessionLifetime and IdleSessionLifetime on the TenantSettings(Base) class.

Fixes #271

### DIFF
--- a/src/Auth0.ManagementApi/Models/TenantSettingsBase.cs
+++ b/src/Auth0.ManagementApi/Models/TenantSettingsBase.cs
@@ -83,13 +83,13 @@ namespace Auth0.ManagementApi.Models
         /// Login session lifetime, how long the session will stay valid (unit: hours).
         /// </summary>
         [JsonProperty("session_lifetime")]
-        public int? SessionLifetime { get; set; }
+        public float? SessionLifetime { get; set; }
 
         /// <summary>
         /// Force a user to login after they have been inactive for the specified number (unit: hours).
         /// </summary>
         [JsonProperty("idle_session_lifetime")]
-        public int? IdleSessionLifetime { get; set; }
+        public float? IdleSessionLifetime { get; set; }
 
         /// <summary>
         /// The selected sandbox version to be used for the extensibility environment.


src/Auth0.ManagementApi/Models/TenantSettingsBase.cs
```diff
@@ -83,13 +83,13 @@ public class TenantSettingsBase
         /// Login session lifetime, how long the session will stay valid (unit: hours).
         /// </summary>
         [JsonProperty("session_lifetime")]
-        public int? SessionLifetime { get; set; }
+        public float? SessionLifetime { get; set; }
```
rcollipi: Would it be an issue to represent these as decimal values rather than as floats?

src/Auth0.ManagementApi/Models/TenantSettingsBase.cs
```diff
@@ -83,13 +83,13 @@ public class TenantSettingsBase
...
 
         /// <summary>
         /// Force a user to login after they have been inactive for the specified number (unit: hours).
         /// </summary>
         [JsonProperty("idle_session_lifetime")]
-        public int? IdleSessionLifetime { get; set; }
+        public float? IdleSessionLifetime { get; set; }
```
rcollipi: Would it be an issue to represent these as decimal values rather than as floats?

src/Auth0.ManagementApi/Models/TenantSettingsBase.cs
damieng: They are IEEE 754 floats on the server side (JavaScript) so you'll only get values that can be contained in floats in .NET too.  There would be no extra precision gained using a Decimal.

src/Auth0.ManagementApi/Models/TenantSettingsBase.cs
damieng: Same response.

src/Auth0.ManagementApi/Models/TenantSettingsBase.cs
rcollipi: Seems reasonable to me!

PR #288, opened 2019-08-21
Prepare 6.5.1 release

damieng:
Fixes get permission calls returning no results on users and roles when paging specified.

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -12,6 +12,9 @@
     <RepositoryType>git</RepositoryType>
     <RepositoryUrl>https://github.com/auth0/auth0.net</RepositoryUrl>
     <PackageReleaseNotes>
+      Version 6.5.1
+      - User and role permissions endpoints in UsersClient and RolesClient paging fix.
+
       Version 6.5.0
       - Assembly is now strong-name-signed so it can be used by other strong-name-signed packages.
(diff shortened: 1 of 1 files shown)


mohit0316:
@damieng  Is this one going out soon?

damieng:
Just published 6.5.1 to NuGet - you should see it now.

To create a new permission you add a ResourceServerScope to the ResourceServer that represents the api (there is unfortunately some discrepancies between the friendly names the dashboard UI of Auth0 uses vs the underlying API names which were based more on the terms the specification uses).

You can see an example of creating a permission and assigning it to a role in this unit test https://github.com/auth0/auth0.net/blob/master/tests/Auth0.ManagementApi.IntegrationTests/RolesTests.cs#L222

mohit0316:
@damieng  Perfect, thanks that's what I was looking for.

mohit0316:
@damieng  This is working fine now. Thank you. As a side note, what would it take to add a `resourceServerId` or `resourceServerIdentifier` to the `UsersClient.GetPermissionsAsync`? I ask since currently the results are returned with all API resource server identifiers for a user's permissions. So, if you are calling this method from an API, it gets results with other resource server identifiers included for a given user. This is for cases where a tenant has two APIs one for stage and other for prod.

Issue #293, opened 2019-08-30
JWT Tokens unable to validate - Invalid KID


williamdenton:
### Description

`Auth0.AuthenticationApi` references `Microsoft.IdentityModel.Protocols.OpenIdConnect v5.4.0` that shipped with a change to the way `kid` is calculated. 

Adding the nuget `Auth0.AuthenticationApi v6.5.2` to any existing service running `AspNet.Security.OpenIdConnect.Server v2.0.0` will perform an implicit upgrade on `Microsoft.IdentityModel.Protocols.OpenIdConnect` which changes the way the kid is generated. This will break clients issued with tokens from the service that use Auth0s supplied javascript library. 
>*note*: this does not apply to tokens issued by auth0

### Prerequisites

- [x] I have checked the [documentation for this library](https://auth0.github.io/auth0.net/).

- [x] I have checked the [Auth0 Community](https://community.auth0.com/) for related posts.

- [x] I have checked for related or duplicate [Issues](https://github.com/auth0/auth0.net/issues) and [PRs](https://github.com/auth0/auth0.net/pulls).

- [x] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md).

- [x] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Environment

Please provide the following:

- Version of Auth0.NET used: `"Auth0.AuthenticationApi" Version="6.5.1"`
- Version of .NET used: `dotnet core 2.1`
- Any other relevant environmental versions:
- Additional libraries that might be affecting your instance:
  * `AspNet.Security.OpenIdConnect.Server` v2.0.0
  * https://www.npmjs.com/package/auth0-js

### Reproduction

`Microsoft.IdentityModel.Protocols.OpenIdConnect` introduced a bug in v`5.4.0` that changed the value for the `kid` (Key ID). Previously the key was set to the thumb print of the certificate. This was fixe in v`5.5.0`

This is the commit that breaks the `keyid` (released in 5.4) https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/760fa3046cc36591c9f7835031745fcd625cf6ec#diff-ad9f507af4d8334a3a5dbcafbb0ded30R53

[...]

PR #295, opened 2019-08-30
Upgrade Microsoft identity library NuGet packages

damieng:
Upgrade to the latest version of the IdentityModel libraries from Microsoft:

- *Microsoft.IdentityModel.Protocols.OpenIdConnect* 5.5.0
- *System.IdentityModel.Tokens.Jwt* 5.5.0


damieng:
Fixes #293

src/Auth0.AuthenticationApi/Auth0.AuthenticationApi.csproj
```diff
@@ -2,9 +2,10 @@
   <Import Project="..\..\build\common.props" />
   <PropertyGroup>
     <AssemblyTitle>Auth0.AuthenticationApi</AssemblyTitle>
-    <TargetFrameworks>netstandard2.0;net452;netstandard1.4;</TargetFrameworks>
     <AssemblyName>Auth0.AuthenticationApi</AssemblyName>
     <PackageId>Auth0.AuthenticationApi</PackageId>
+    <TargetFrameworks>netstandard2.0;net452;netstandard1.4;</TargetFrameworks>
+    <NetStandardImplicitPackageVersion>2.0.3</NetStandardImplicitPackageVersion>
```
damieng: This was unfortunately required due to some internal housekeeping in System.IdentityModel.Tokens.Jwt that broke things without it in a minor.

williamdenton:
Please ensure the javascript library is also fixed else you will break people

damieng:
I thought 5.5 reverted the behavior to the same as 5.3?  Is it instead still subtly different?

williamdenton:
for users happily running on the current version referencing 5.4, upgrading to 5.5 is a breaking change. Its breaking because Auth0's javascript can't validate the jwt if the kid changes.


you need to fix this issue in your javascript library https://github.com/auth0/node-jwks-rsa/issues/5

williamdenton:
to resolve this for us we had to manually patch the js to this


```js
instance.getSigningKey = function (kid, cb) {
    instance.logger(`Fetching signing key for '${kid}'`);

     instance.getSigningKeys((err, keys) => {
      if (err) {
        return cb(err);
      }

      const key = keys.find((k) => k.x5t === kid || k.kid === kid);
      if (key) {
        return cb(null, key);
      }
      instance.logger(`Unable to find a signing key that matches '${kid}'`);
      return cb(new JwksClientOriginal.SigningKeyNotFoundError(`Unable to find a signing key that matches '${kid}'`));
    });
  };
```


the crucial line is
```js
const key = keys.find((k) => k.x5t === kid || k.kid === kid);
```

which is changed from
```js
const key = keys.find(k => k.kid === kid); 
```

damieng:
I understand the issue you've linked to in isolation - I just don't see how that library would ever be attempting to work with a JWT that came from Auth0.Net - this library only checks the token it is given - it does not create ones that would be consumed elsewhere.

williamdenton:
our problem stemmed from including this library on our STS server which is currently issuing tokens. We added this library to our service as we are in the process of migrating to auth0, but are still issuing jwts to our own clients.

The effect of adding it to our existing server meant that the `/well-known/jwks` was not usable by the javascript client.

Its a fairly niche edge case but its impact is quite large.

if you have other customers in a similar situation but who are stable on version 5.4, upgrading them to 5.5 will break all clients with currently valid jwts being validated by your javascript library.

damieng:
@williamdenton I think that is quite an unusual edge case. I really want to get this update out as it includes some other fixes and could prevent people falling into the 5.4 trap as well.

I'll put a note in the changelog and try and figure out how we can move node-jwks-rsa forward - there's a whole lot of issues outstanding on that repo not least of all the most recent version seems to have introduced an unintentional breaking change.

PR #301, opened 2019-09-13
Add scope parameter to token request for client credentials

Boomit2016:


damieng:
That method maps to this Auth0 endpoint https://auth0.com/docs/api/authentication#client-credentials-flow which does not take a scope parameter.

Issue #250, opened 2019-04-05
How to test this package in 6.X+?


carusology:
### Description

We've been using the `5.X` versions of this package for some time as part of our user management workflows in our ASP.NET Core v2.1 product. We have built unit tests over those libraries by mocking the various `I*Client` interfaces such as `IConnectionsClient`, `IUsersClient`, etc. Now I see the interfaces are retired, and I don't understand why. 

The [nuget page](https://www.nuget.org/packages/Auth0.ManagementApi/) states that this is so "All I*Client interfaces have been removed so adding endpoints is no longer breaking." There is a link to a migration guide (a dead link, but I assume is supposed to point [here](http://auth0.github.io/auth0.net/migrating.html#migrating-from-version-5-to-version-6)?), which states a surprising comment: " As this library is a concrete implementation of calling the Auth0 management APIs and not a general abstraction layer over management this interface served no purpose." I have to respectfully disagree here. Us, the consumers of your API, use and wrap these interfaces with composition to alter behavior, set default values from configuration, manage rate-limiting, add logging, etc. Remove the interface, and you remove our ability to compose these base implementations.

The end result of that decision is now I'm looking at this very major change and I'm left scratching my head with a straightforward question: **How are we intended to write tests for this code?**

If we are to avoid unit tests in favor of integration tests, Auth0 is rate-limited, so the tests will take forever to run and we have to keep track of how many requests are going out a second. This kind of tracking as a side-effect is pretty easily done with composition, but composition in .NET is traditionally done via interface. We also can't use tools such as Moq to mock the interface because, again, no interface exists. **Are we intended to use partial mocks? Is that how it was envisioned that we would write tests?**

For what we should do on our end, I'm left thinking we should locally reimplement the interfaces that were just removed and use these various `*Clients` classes as the base implementations of those interfaces. But if we're doing that, it goes back to my original confusion as to why these interfaces were removed in the first place, and how the maintainers expect us to test the various clients in our own code bases. Maybe this _is_ the intention? Should we write our own interfaces around the functionally we use?

Anyway, I'd love some insight into how consumers should be thinking about their testing strategies for this library as a result of this change.

### Prerequisites

[X] I have checked the [documentation for this library](https://auth0.github.io/auth0.net/).

[X] I have checked the [Auth0 Community](https://community.auth0.com/) for related posts.

[...]

PR #306, opened 2019-10-14
Accept IApiConnection on AuthenticationApiClient & ManagementApiClient

damieng:
Provides a simple way to mock out the AuthenticationApiClient and ManagementApiClient underlying implementations without sprawling concerns or breaking changes when we add endpoints.

Related to #250


src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
```diff
@@ -406,11 +411,24 @@ public async Task UnlinkUserAsync(UnlinkUserRequest request)
         }
 
         /// <summary>
-        /// Disposes of any owned disposable resources such as the ApiConnection.
+        /// Dispose of any managed resources such as the <see cref="IApiConnection"/>.
+        /// </summary>
+        /// <param name="disposing">Whether to dispose of managed resources right now.</param>
+        protected virtual void Dispose(bool disposing)
```
joshcanhelp: I'm confused why this is necessary. Why would you call this with `disposing` set to false?

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
```diff
@@ -17,10 +17,17 @@ namespace Auth0.AuthenticationApi
     public class AuthenticationApiClient : IAuthenticationApiClient, IDisposable
     {
         private readonly Uri _baseUri;
-        private readonly ApiConnection _apiConnection;
+        private IApiConnection _apiConnection;
```
joshcanhelp: Does `ApiConnection` extend `IApiConnection`? Or am I missing a different relationship here?

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
damieng: It's the standard .NET pattern for disposing objects that can be inherited from.  VS suggests it strongly from the IDE.

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
damieng: IApiConnection is the interface, ApiConnection is our concrete class.  By accepting an interface instead customers can create their own implementation that instead returns canned responses for their testing.

damieng:
Going to do more work on this PR splitting `IApiConnection` to `IManagementApiConnection` and `IAuthenticationApiConnection` so that either side is easier  to mock and also combine the various overloads for different http methods into a single one with a parameter to further aid with mocking.

damieng:
IApiConnection is a bad interface for AuthenticationClient, there should be two interfaces one for each. Rejecting in favor of an alternative (incoming).

PR #312, opened 2019-11-12
Replace ApiData with RateLimitException

damieng:
Right now ApiData is the only part of the `MangementApiClient` and `AuthenticationApiClient` that is not thread safe.

Our Java SDK takes a different approach and only exposes the rate limits on failure https://github.com/auth0/auth0-java/pull/153

This PR brings this SDK in-line with that in order to better improve scalability by making the clients totally thread-safe so they can easily re-use the `HttpClient` without the developer needing to jump through hoops.

The implementation requires that a new `RateLimitException` is introduced. To ensure this is caught under existing code paths for consumers it inherits from `ApiException`.

Because that class has metadata that is not relevant to Rate Limits such as `ApiError` what was effectively `ApiException` becomes `ErrorApiException` which also inherits from ApiException.  Most cases would be caught although some might need to change their catching of `ApiException` to `ErrorApiException` in order to access ApiData in their handler.


src/Auth0.Core/ApiError.cs
```diff
@@ -27,10 +28,23 @@ public class ApiError
...
 
-        /// <summary>
-        /// The HTTP Status code.
-        /// </summary>
-        [JsonProperty("statusCode")]
-        public int StatusCode { get; set; }
+        public static ApiError ParseOrDefault(string responseContent)
+        {
+            if (String.IsNullOrEmpty(responseContent))
+                return default;
```
joshcanhelp: **Question:** what does `return default` do?

src/Auth0.Core/ApiError.cs
```diff
@@ -27,10 +28,23 @@ public class ApiError
...
+
+            try
+            {
+                return JsonConvert.DeserializeObject<ApiError>(responseContent);
```
joshcanhelp: Is this an API error from our server? Is it expecting a specific format for this JSON?

src/Auth0.Core/ApiError.cs
```diff
@@ -27,10 +28,23 @@ public class ApiError
...
+            }
+            catch (Exception)
+            {
+                return new ApiError
+                {
+                    Error = responseContent,
```
joshcanhelp: Does `Error` represent an error code? If so, I would prefer `unknown` or thereabouts instead of a duplicate of the message.

src/Auth0.Core/Exceptions/ErrorApiException.cs
```diff
@@ -0,0 +1,62 @@
...
+            : base(message)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ErrorApiException"/> class with a specified error message
+        /// and a reference to the inner exception that is the cause of this exception.
+        /// </summary>
+        /// <param name="message">The error message that explains the reason for the exception.</param>
+        /// <param name="innerException">The exception that is the cause of the current exception, or a null
```
joshcanhelp: **Question:** Can you explain an "inner" exception? Like a child of the main exception?

src/Auth0.Core/Exceptions/ErrorApiException.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
+using System.Net;
+
+namespace Auth0.Core.Exceptions
+{
+    /// <summary>
+    /// Represents errors that occur when making API calls.
+    /// </summary>
+    public class ErrorApiException : ApiException
```
joshcanhelp: So this would be used for all errors besides a rate limit encountered? And then we could add additional exceptions for other errors, if we're able to do something helpful?

src/Auth0.Core/Exceptions/RateLimitApiException.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RateLimitApiException"/> class with a specified error message
+        /// and a reference to the inner exception that is the cause of this exception.
+        /// </summary>
+        /// <param name="message">The error message that explains the reason for the exception.</param>
+        /// <param name="innerException">The exception that is the cause of the current exception, or a null 
+        /// reference if no inner exception is specified.</param>
+        public RateLimitApiException(string message, Exception innerException)
```
joshcanhelp: What kind of inner exception would there be for a rate limit hit?

src/Auth0.Core/Exceptions/RateLimitApiException.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+        }
+
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RateLimitApiException"/> class with a specified <paramref name="rateLimit"/>.
+        /// </summary>
+        /// <param name="rateLimit"><see cref="Http.RateLimit"/> received on the API call that failed.</param>
+        public RateLimitApiException(RateLimit rateLimit)
+            : this("Rate limits exceeded")
```
joshcanhelp: Here's the actual error messaging used from the server:

```
{"statusCode":429,"error":"Too Many Requests","message":"Global limit has been reached","errorCode":"too_many_requests"}
```

src/Auth0.Core/Http/RateLimit.cs
```diff
@@ -21,5 +23,34 @@ public class RateLimit
         /// The date and time offset at which the current rate limit window is reset.
         /// </summary>
         public DateTimeOffset Reset { get; internal set; }
+
+        public static RateLimit Parse(HttpResponseHeaders headers)
```
joshcanhelp: Is this available on all requests? Not on a rate limit hit only?

src/Auth0.Core/ApiError.cs
damieng: It's the same as `return null` for reference types and `return new x` for value types.  Given ApiError is a reference type I could just put `return null` here - I briefly toyed with making ApiError a struct hence the default here which works for both.

src/Auth0.Core/ApiError.cs
damieng: There is an ApiError deserializer that was already part of the code base to create ApiErrors from our responses.

src/Auth0.Core/ApiError.cs
damieng: That's what was already here - just moved around.  We could break it if you think there's value.

src/Auth0.Core/Exceptions/ErrorApiException.cs
damieng: Yeah it's what exception caused this exception. These are just standard Exception constructors every constructor in .net should have.

src/Auth0.Core/Exceptions/ErrorApiException.cs
damieng: Yep.  Previously there was only one exception type which had ApiError.  Now we have RateLimitException which won't have that so I moved ApiError to another new sub-type of ApiException.

This is the safest path for migration - if they were using ApiError they'll just need to switch their catch from ApiException to ErrorApiException when they see the problem in the compiler.

If I was to leave the existing ApiException as-is then RateLimitException would be not handled and they could instead see exceptions in production they previously handled when those 429 errors were just plain ApiException.

src/Auth0.Core/Exceptions/RateLimitApiException.cs
damieng: Hmm, not sure if we want to propagate the server message into an actual exception message, let me think that over.

src/Auth0.Core/Http/RateLimit.cs
damieng: It is, but we will no longer expose it unless it's an error.  Realistically there's not much you can do with it as it's a global rate. You can't co-ordinate between various clients to stay under.  If it were a per-instance rate limit it might make sense for simple single-threaded apps to expose it.

src/Auth0.Core/Exceptions/RateLimitApiException.cs
joshcanhelp: Maybe not propagate directly but maybe use the same text hard-coded here.

src/Auth0.Core/Http/RateLimit.cs
joshcanhelp: > It is, but we will no longer expose it unless it's an error.

If this is only exposed in an error then the amount remaining and the limit itself is irrelevant. 

> You can't co-ordinate between various clients to stay under.

If you see the remaining/reset before, you could pause, if that made sense for the app.

src/Auth0.Core/ApiError.cs
joshcanhelp: I'm just thinking that duplicating the content in both is unnecessary. An unset `Error` would be more helpful than the same as `Message`

src/Auth0.Core/Http/RateLimit.cs
damieng: But only if it was the only instance of the app running anywhere.  You can't tell what the other clients are going to do and it's a per-tenant rate.

src/Auth0.Core/Exceptions/RateLimitApiException.cs
damieng: None that I can think of.  You just have to have these constructors for Exceptions in .NET. It's part of the implied contract.

src/Auth0.Core/Exceptions/RateLimitApiException.cs
damieng: Will pass actual error in another PR.

src/Auth0.Core/ApiError.cs
damieng: Will figure out a better pattern for this in another PR.

PR #314, opened 2019-11-13
Alternative connection system

damieng:
The current connection system suffers from various problems, specifically `IApiConnection`:

- Has an unnecessarily large surface area in terms of methods and parameters
- This makes it difficult to mock
- Is not accepted by the ApiClients so you can't mock anyway
- Has mixed concerns - both building URLs and executing requests
- Mixes scenarios - Auth and Management have different requirements
- Rate limits are not concurrency aware see https://github.com/auth0/auth0.net/pull/312

This PR replaces `IApiConnection` with `IAuthenticationConnection` and `IManagementConnection` that address these issues.

It also moves some files from Auth0.Core to Authentication and Management that are not used by the other library.  A long-term goal might be to remove Auth0.Core entirely.


damieng:
Will address #250

damieng:
Closing in favor of smaller commits with Auth and Management tackled independently. (same design however)

PR #317, opened 2019-11-20
Root profile fields on signup

damieng:
Added root profile attributes to the signup request and response.

### DIFF
--- a/tests/Auth0.AuthenticationApi.IntegrationTests/DatabaseConnectionTests.cs
+++ b/tests/Auth0.AuthenticationApi.IntegrationTests/DatabaseConnectionTests.cs
@@ -1,11 +1,10 @@
-﻿using System;
-using System.Threading.Tasks;
-using Auth0.AuthenticationApi.Models;
-using Auth0.Core.Exceptions;
+﻿using Auth0.AuthenticationApi.Models;
 using Auth0.ManagementApi;
 using Auth0.ManagementApi.Models;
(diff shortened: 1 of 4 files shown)


tests/Auth0.AuthenticationApi.IntegrationTests/DatabaseConnectionTests.cs
```diff
@@ -63,17 +68,11 @@ public async Task Can_signup_user_and_change_password()
             signupUserResponse.Id.Should().NotBeNull();
             signupUserResponse.EmailVerified.Should().BeFalse();
             signupUserResponse.Email.Should().Be(signupUserRequest.Email);
-
-            // Change the user's Email address
-            var changePasswordRequest = new ChangePasswordRequest
```
damieng: This test made no sense... password isn't even supposed to be on that request so now it's moved from that request object there's nothing to test.

lbalmaceda: @damieng This was allowed previously. Probably now deprecated and allowed only with a legacy flag on the account. AFAIK today if you send the password, it will be ignored by the server but the request won't fail.

damieng: Right, that's good context.  This test as it was was making the SDK more confusing to users in order to test that a server rejects things so I'm even more 👍 with this change.

PR #318, opened 2019-11-20
WithConnectionScope method added to AuthorizationUrlBuilder

flarocca:
### Changes

Adding **WithConnectionScope** to **AuthorizationUrlBuilder **

Actually, the AuthorizationUrlBuilder does not offer any way to include Connection Scopes when building the authorization URL. This forces us to concatenate the aforementioned field manually, e.g.:

```
using (var authenticationApiClient = new AuthenticationApiClient(new Uri("BASE_URL")))
{
    var urlBuilder = authenticationApiClient.BuildAuthorizationUrl();
    urlBuilder
        // More stuff...
        .WithScope("offline_access openid");

    var url = urlBuilder.Build();

    return url.AbsoluteUri + "&connection_scope=IDP scopes";
}
```

By introducing this new method *WithConnectionScope*, we will be able to include connection scopes following the same fluent-api approach the builder already uses, resulting in a cleaner solution. E.g.:

```
using (var authenticationApiClient = new AuthenticationApiClient(new Uri("BASE_URL")))
{
    var urlBuilder = authenticationApiClient.BuildAuthorizationUrl();
    urlBuilder
        .WithResponseType(AuthorizationResponseType.Code)
        // More stuff...
        .WithConnectionScope("IDP scopes");

    var url = urlBuilder.Build();

    return url.AbsoluteUri;
}
```

### References

N/A

### Testing

This new method can be tested by calling **WithConnectionScope** method on the **AuthorizationUrlBuilder** with the scopes you want to pass to the IDP. A new query string parameter will be included in the constructed URL, e.g.:

https://auth0-dotnet-integration-tests.auth0.com/authorize?other_parameters&connection_scope=ConnectionScope

### Checklist

### DIFF
--- a/src/Auth0.AuthenticationApi/Builders/AuthorizationUrlBuilder.cs
+++ b/src/Auth0.AuthenticationApi/Builders/AuthorizationUrlBuilder.cs
@@ -137,5 +137,17 @@ namespace Auth0.AuthenticationApi.Builders
 
             return this;
         }
+
+        /// <summary>
+        /// Adds a new connection_scope query string parameter.
+        /// </summary>
+        /// <param name="connectionScope">The connection scope to be passed to the corresponding connection. Multiple scopes must be separated by a space character.</param>
+        /// <returns>The <see cref="AuthorizationUrlBuilder"/>.</returns>
+        public AuthorizationUrlBuilder WithConnectionScope(string connectionScope)
+        {
+            AddQueryString("connection_scope", connectionScope);
+
+            return this;
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


damieng:
Hi! Thanks for the PR.

I wonder if this should be "AddConnectionScopes" and have a params array it joins with commas internally?

Would perhaps be clearer to users it can take multiple scopes and mean they don't have to check the docs to see what the separator is.

flarocca:
@damieng Thanks for taking the time to review the PR and suggesting alternatives. I really appreciate it!
Even though it sounds like a very good idea, for the sake of consistency, I believe we should keep `With` prefix and a single string as the input parameter.
Actually, there is a `WithScope` method receiving a single `string` parameter as input:

```
/// <summary>
/// Adds a new scope query string parameter.
/// </summary>
/// <param name="scope">The scope. Multiple scopes must be separated by a space character.</param>
/// <returns>The <see cref="AuthorizationUrlBuilder"/>.</returns>
public AuthorizationUrlBuilder WithScope(string scope)
{
    AddQueryString("scope", scope);

    return this;
}
```

So changing the prefix to `Add` would break consistency, in my opinion.

Having said that, do you still think I should rename it as `AddScope` and change the input parameter to `string[] params`?

Thanks a lot.

PR #321, opened 2019-11-21
Replaces IApiConnection with IAuthenticationConnection

damieng:
Replaces the IApiConnection with IAuthenticationConnection and ApiConnection with HttpClientAuthenticationConnection in order to:

1. Improve testability
2. Separate Auth and Management functionality
3. Clean up documentation

### DIFF
--- a/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
+++ b/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
@@ -1,7 +1,6 @@
 ﻿using Auth0.AuthenticationApi.Builders;
 using Auth0.AuthenticationApi.Models;
 using Auth0.AuthenticationApi.Tokens;
-using Auth0.Core.Http;
 using System;
 using System.Collections.Generic;
 using System.Net.Http;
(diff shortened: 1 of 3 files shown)


src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
```diff
@@ -10,199 +9,185 @@
 namespace Auth0.AuthenticationApi
 {
     /// <summary>
-    /// Client for communicating with the Auth0 Authentication API.
+    /// Client for the Auth0 Authentication API.
     /// </summary>
     /// <remarks>
-    /// Full documentation for the Authentication API is available at https://auth0.com/docs/auth-api
+    /// Full documentation on the Auth0 Authentication API is available at https://auth0.com/docs/auth-api
```
joshcanhelp:
```suggestion
    /// Full documentation on the Auth0 Authentication API is available at https://auth0.com/docs/api/authentication
```

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
```diff
@@ -10,199 +9,185 @@
...
-        public AuthenticationApiClient(string domain, HttpMessageHandler handler = null)
-            : this(new Uri($"https://{domain}"), handler)
+        /// <param name="handler">Optional <see cref="IAuthenticationConnection"/> used to influence connection behavior.
+        /// Defaults to a freshly created <see cref="HttpClientAuthenticationConnection"/> that uses a single <see cref="HttpClient"/>.</param>
+        /// <remarks>To use a custom <see cref="HttpClient"/> or <see cref="HttpMessageHandler"/> create a 
+        /// <see cref="HttpClientAuthenticationConnection"/> passing that into the constructor. e.g.
+        /// <code>var client = new AuthenticationApiClient(new HttpClientAuthenticationConnection(myHttpClient));</code> or
+        /// <code>var client = new AuthenticationApiClient(new HttpClientAuthenticationConnection(myHttpMessageHandler));</code> or
+        /// </remarks>
+        public AuthenticationApiClient(Uri baseUri, IAuthenticationConnection connection = null)
```
joshcanhelp: Why is this using `baseUri` and the next one is using `domain`?

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
```diff
@@ -211,134 +196,118 @@ public async Task<AccessTokenResponse> GetTokenAsync(AuthorizationCodePkceTokenR
...
+        /// <returns><see cref="AccessTokenResponse"/> containing the desired token information.</returns>
         public async Task<AccessTokenResponse> GetTokenAsync(RefreshTokenRequest request)
         {
-            var parameters = new Dictionary<string, object> {
+            var body = new Dictionary<string, string> {
                 { "grant_type", "refresh_token" },
-                { "refresh_token", request.RefreshToken },
                 { "client_id", request.ClientId },
-                { "client_secret", request.ClientSecret }
+                { "client_secret", request.ClientSecret },
```
joshcanhelp: You may want to make this optional. If this library uses PKCE, that's an indication it might not be storing/using a secret. Refresh tokens can be exchanged without one (if the application is setup to do that).

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
```diff
@@ -211,134 +196,118 @@ public async Task<AccessTokenResponse> GetTokenAsync(AuthorizationCodePkceTokenR
...
+                { "refresh_token", request.RefreshToken }
             };
 
-            if (!string.IsNullOrEmpty(request.Audience))
-            {
-                parameters.Add("audience", request.Audience);
-            }
+            body.AddIfNotEmpty("audience", request.Audience);
+            body.AddIfNotEmpty("scope", request.Scope);
```
joshcanhelp: You never need to send these. The refresh token is tied to a specific audience and scope:

https://auth0.com/docs/api/authentication#refresh-token

src/Auth0.AuthenticationApi/IAuthenticationConnection.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Auth0.AuthenticationApi
+{
+    /// <summary>
+    /// Interface used to customize implementation or mock behavior of <see cref="AuthenticationApiClient"/>.
+    /// </summary>
+    public interface IAuthenticationConnection
```
joshcanhelp: 👍 this is nice!

damieng:
> I feel like we've covered this ground before but ... tests?

I'm not adding any new functionality so the existing integration tests should have everything covered.

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
damieng: So you can either pass in a whole url, e.g.

 ```csharp
var x = new AuthenticationApiClient("https://damieng.auth0.com")
```

or just your domain name

```csharp
var x = new AuthenticationApiClient("damieng.auth0.com");
```

It was already part of the existing design and I didn't have a problem with it.

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
damieng: So just add client_secret conditionally?

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
damieng: Okay, I could remove them but that's going beyond the scope of this PR.

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
joshcanhelp: Yes

damieng: Let's do this and the removal of the unnecessary refresh token parameters in a separate PR to keep the concerns separated.  I'll put it out this morning once this one is in.

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
joshcanhelp: Ok, I understand this better. It was unclear at first that it was only the first type that was different.

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
joshcanhelp: It looks like you're re-writing quite a few of these methods and this could be part of that cleanup. Sending or not sending does not change the behavior of this endpoint, it would be like sending a randomly-generated parameter and value. 

Up to you, it won't hold up my approval.

joshcanhelp:
@damieng - RE: tests ... ok, that makes sense. I assume it covers, implicitly, the new class as well?

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
damieng: I'd rather do it as a separate PR - that way I can refer to the change separately in the change log. From an outside perspective they are two different things.

damieng:
> @damieng - RE: tests ... ok, that makes sense. I assume it covers, implicitly, the new class as well?

That is correct.

PR #323, opened 2019-11-22
Improve API builders

damieng:
Improves the URL builders by:

1. Improving the doc comments everywhere including adding links to online Auth API ref
2. Adding `Url` class alternatives to the `string` ones that take URLs (constructors and methods)
3. Switches from `AddQueryValue` + `return` to just `WithValue` for shorter code
4. Adds methods that take arrays of parameters for things like Scopes so users don't need to know separator rules
5. Removes the `WithAccessToken` method from AuthorizationBuilder... no idea what this was supposed to be for but it's not part of what's accepted.

### DIFF
--- a/src/Auth0.AuthenticationApi/Builders/AuthorizationUrlBuilder.cs
+++ b/src/Auth0.AuthenticationApi/Builders/AuthorizationUrlBuilder.cs
@@ -1,153 +1,165 @@
 using Auth0.AuthenticationApi.Models;
+using System;
 using System.Linq;
 
 namespace Auth0.AuthenticationApi.Builders
 {
     /// <summary>
-    /// Used to build am authorization URL.
+    /// Builder class used to fluently construct an authorization URL.
     /// </summary>
+    /// <remarks>
+    /// See https://auth0.com/docs/api/authentication#login for more details.
+    /// </remarks>
     public class AuthorizationUrlBuilder : UrlBuilderBase<AuthorizationUrlBuilder>
     {
         /// <summary>
         /// Initializes a new instance of the <see cref="AuthorizationUrlBuilder"/> class.
         /// </summary>
-        /// <param name="baseUrl">The base URL of the Authentication API.</param>
-        public AuthorizationUrlBuilder(string baseUrl) 
+        /// <param name="baseUrl">Base URL of the Authentication API represented as a <see cref="String"/>.</param>
+        public AuthorizationUrlBuilder(string baseUrl)
             : base(baseUrl, "authorize")
         {
         }
 
         /// <summary>
-        /// Adds an access_token query string parameter.
+        /// Initializes a new instance of the <see cref="AuthorizationUrlBuilder"/> class.
         /// </summary>
-        /// <param name="accessToken"></param>
-        /// <returns></returns>
-        public AuthorizationUrlBuilder WithAccessToken(string accessToken)
+        /// <param name="baseUrl">Base URL of the Authentication API represented as a <see cref="Uri"/>.</param>
+        public AuthorizationUrlBuilder(Uri baseUrl)
+            : base(baseUrl, "authorize")
         {
-            AddQueryString("access_token", accessToken);
-
-            return this;
         }
 
         /// <summary>
-        /// Adds a new client_id query string parameter.
+        /// Adds the `client_id` query string parameter specifying the Client ID of the application.
         /// </summary>
-        /// <param name="clientId">The client identifier.</param>
-        /// <returns>The <see cref="AuthorizationUrlBuilder"/>.</returns>
+        /// <param name="clientId">Client ID of the application.</param>
+        /// <returns>Current <see cref="AuthorizationUrlBuilder"/> to allow fluent configuration.</returns>
         public AuthorizationUrlBuilder WithClient(string clientId)
         {
-            AddQueryString("client_id", clientId);
-
-            return this;
+            return WithValue("client_id", clientId);
         }
 
         /// <summary>
-        /// Add a new connection query string parameter.
+        /// Adds the `connection` query string parameter specifying the connection name.
         /// </summary>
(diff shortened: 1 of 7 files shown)


src/Auth0.AuthenticationApi/Builders/AuthorizationUrlBuilder.cs
```diff
@@ -1,153 +1,165 @@
...
-            return this;
+        /// <summary>
+        /// Adds the `connection_scope` query string parameter.
+        /// </summary>
+        /// <param name="connectionScope">Connection scopes to be passed to the corresponding connection.</param>
+        /// <returns>Current <see cref="AuthorizationUrlBuilder"/> to allow fluent configuration.</returns>
+        public AuthorizationUrlBuilder WithConnectionScopes(params string[] connectionScope)
+        {
+            return WithConnectionScope(String.Join(" ", connectionScope));
         }
```
lbalmaceda: `max_age` setter missing. Applies to this or not?

src/Auth0.AuthenticationApi/Builders/LogoutUrlBuilder.cs
```diff
@@ -1,54 +1,70 @@
...
         /// <summary>
-        /// Adds a client_id query string parameter.
+        /// Adds the `redirect_uri` query string parameter specifying the redirect URI.
         /// </summary>
-        /// <param name="clientId">The URL.</param>
-        /// <returns>The <see cref="LogoutUrlBuilder"/>.</returns>
-        public LogoutUrlBuilder WithClientId(string clientId)
+        /// <param name="uri"><see cref="Uri"/> to redirect to.</param>
+        /// <returns>Current <see cref="LogoutUrlBuilder"/> to allow fluent configuration.</returns>
+        public LogoutUrlBuilder WithReturnUrl(Uri uri)
```
lbalmaceda: any reason to add this new `Uri` variants? and, do we want to deprecate the `string` or keep both?

src/Auth0.AuthenticationApi/Builders/LogoutUrlBuilder.cs
```diff
@@ -1,54 +1,70 @@
...
+        /// Adds the `federated` flag query string parameter (no value).
         /// </summary>
-        /// <returns>The <see cref="LogoutUrlBuilder"/>.</returns>
+        /// <returns>Current <see cref="LogoutUrlBuilder"/> to allow fluent configuration.</returns>
         public LogoutUrlBuilder Federated()
         {
-            AddQueryString("federated", null);
-
-            return this;
+            return WithValue("federated", null);
```
lbalmaceda: will a `null` value mean the query param is included in the final string url that gets constructed? if not, you could set it to `1` to trigger that

src/Auth0.AuthenticationApi/Builders/SamlUrlBuilder.cs
```diff
@@ -1,64 +1,74 @@
...
         /// Initializes a new instance of the <see cref="SamlUrlBuilder"/> class.
         /// </summary>
-        /// <param name="baseUrl">The base URL.</param>
-        /// <param name="client">The client id.</param>
-        public SamlUrlBuilder(string baseUrl, string client)
-            :base(baseUrl, "samlp/{client}")
+        /// <param name="baseUrl">Base URL of the Authentication API represented as a <see cref="String"/>.</param>
+        /// <param name="clientId">Client ID of the application.</param>
+        public SamlUrlBuilder(string baseUrl, string clientId)
+            : base(baseUrl, "samlp/{client}")
```
lbalmaceda: won't it be more readable if this changes to
```suggestion
            : base(baseUrl, "samlp/{clientId}")
```

and below you keep "setting" the segment `clientId` as well?

(there are a few places in this diff that this comment applies)

tests/Auth0.AuthenticationApi.IntegrationTests/UriBuildersTests.cs
```diff
@@ -97,6 +114,20 @@ public void Can_build_logout_url_with_return_url()
...
         }
 
+        [Fact]
+        public void Can_build_logout_url_with_return_url_as_uri()
+        {
+            var authenticationApiClient = new AuthenticationApiClient(GetVariable("AUTH0_AUTHENTICATION_API_URL"));
+
+            var logoutUrl = authenticationApiClient.BuildLogoutUrl()
+                .WithReturnUrl(new Uri("http://www.jerriepelser.com/test"))
+                .Build();
```
lbalmaceda: please include in the tests the federated value

src/Auth0.AuthenticationApi/Builders/AuthorizationUrlBuilder.cs
damieng: It wasn't there before do you think there is value?

src/Auth0.AuthenticationApi/Builders/LogoutUrlBuilder.cs
damieng: .NET has unfortunately used both everywhere so doing this to keep it familiar.  Because they have the same method name it just works as needed no matter whether they have a string or a Uri object.

src/Auth0.AuthenticationApi/Builders/LogoutUrlBuilder.cs
damieng: It just means it will add `?federated` without the `=` part which any other value would cause.

src/Auth0.AuthenticationApi/Builders/AuthorizationUrlBuilder.cs
damieng: Let's add it in a future update/PR - nobody has asked for it yet and it wouldn't be a breaking change to add it later.

PR #326, opened 2019-11-22
Implementation HS256 signature verification

damieng:
Implements (but does not yet wire up) the `HS256` signature verification.


src/Auth0.AuthenticationApi/Tokens/IdTokenValidator.cs
```diff
@@ -33,9 +33,8 @@ internal static async Task AssertTokenMeetsRequirements(this IdTokenRequirements
 
             var token = DecodeToken(rawIDToken);
 
-            // For now we want to support HS256 + ClientSecret as we just had a major release.
-            // TODO: In the next major (v4.0) we should remove this condition as well as Auth0ClientOptions.ClientSecret
```
lbalmaceda: I'm confused with this comment about v4 and the one below for v7. Should we keep supporting client_secret (and hs256) or not?

src/Auth0.AuthenticationApi/Tokens/IdTokenValidator.cs
```diff
@@ -33,9 +33,8 @@ internal static async Task AssertTokenMeetsRequirements(this IdTokenRequirements
...
-            if (token.SignatureAlgorithm != "HS256")
+            // TODO: Make signatureVerifier non-optional and not skipped for 'HS256' in 7.0.0
+            if (token.SignatureAlgorithm != "HS256" || signatureVerifier != null)
```
lbalmaceda: where is `signatureVerifier` set and on which scenarios? and more importantly what if the token is signed with `alg=none`?

src/Auth0.AuthenticationApi/Tokens/SymmetricSignatureVerifier.cs
```diff
@@ -0,0 +1,69 @@
...
+    internal class SymmetricSignatureVerifier : ISignatureVerifier
+    {
+        private readonly SecurityKey signingKey;
+
+        public SymmetricSignatureVerifier(SecurityKey signingKey)
+        {
+            this.signingKey = signingKey;
+        }
+
+        public static SymmetricSignatureVerifier FromClientSecret(string clientSecret, bool isBase64Encoded = false)
```
lbalmaceda: `isBase64Encoded ` is legacy (+3 years old setting). Please remove if this hits a new major.

src/Auth0.AuthenticationApi/Tokens/SymmetricSignatureVerifier.cs
```diff
@@ -0,0 +1,69 @@
...
+                    ValidateIssuerSigningKey = true,
+                    IssuerSigningKey = signingKey
+                };
+
+                securityTokenHandler.ValidateToken(token, validationParameters, out var verifiedToken);
+                return verifiedToken;
+            }
+            catch (SecurityTokenException e)
+            {
+                throw new IdTokenValidationException("Invalid ID token signature.", e);
```
lbalmaceda: According to the docs, the error message should be

```suggestion
                throw new IdTokenValidationException("Invalid token signature.", e);
```

tests/Auth0.AuthenticationApi.IntegrationTests/Tokens/IdTokenValidatorTests.cs
```diff
@@ -61,31 +61,45 @@ public async void ThrowsWhenIdTokenCanNotBeDecoded()
         }
 
         [Fact]
```
lbalmaceda: Please add a test to cover the `none` algorithm case

src/Auth0.AuthenticationApi/Tokens/IdTokenValidator.cs
damieng: This was a hold-over from borrowing the code from auth0-oidc-client-net and can be ignored.

tests/Auth0.AuthenticationApi.IntegrationTests/Tokens/IdTokenValidatorTests.cs
damieng: That test already exists here https://github.com/auth0/auth0.net/blob/master/tests/Auth0.AuthenticationApi.IntegrationTests/Tokens/IdTokenValidatorTests.cs#L93

src/Auth0.AuthenticationApi/Tokens/IdTokenValidator.cs
damieng: That will be in the next PR that actually allows the HS256 validation to be used.

PR #328, opened 2019-11-24
Remove .NET Standard 1.4 build

damieng:
Remove .NET Standard 1.4 builds - too many Snyk complaints about HttpClient in that version as well as some missing functionality in forthcoming PRs.


damieng:
Yes, this will be part of 7.0.0.  I've gone ahead and created a milestone and retroactively applied the PRs and issues that will be part of it.

The Snyk errors here will go away I think after we merge and then remove them from Snyk - it gets very upset when build targets are removed.

PR #330, opened 2019-11-25
Upgrade test dependencies

damieng:


damieng:
Weird issue upgrading to latest Microsoft Test SDK despite not being a major.  Don't have cycles to investigate it.

PR #332, opened 2019-11-25
Switch management connection to interface from concrete type

damieng:
Management API was using a mix of interfaces and the concrete type for the connection. 

Normalizes on the interface.

### DIFF
--- a/src/Auth0.ManagementApi/ManagementApiClient.cs
+++ b/src/Auth0.ManagementApi/ManagementApiClient.cs
@@ -10,8 +10,8 @@ namespace Auth0.ManagementApi
     /// </summary>
     public class ManagementApiClient : IDisposable
     {
-        private readonly ApiConnection _apiConnection;
-        private bool disposed;
+        readonly IApiConnection connection;
+        bool disposed;
 
         /// <summary>
         /// Contains all the methods to call the /blacklists/tokens endpoints.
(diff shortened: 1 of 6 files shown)


src/Auth0.ManagementApi/ManagementApiClient.cs
```diff
@@ -10,8 +10,8 @@ namespace Auth0.ManagementApi
     /// </summary>
     public class ManagementApiClient : IDisposable
     {
-        private readonly ApiConnection _apiConnection;
-        private bool disposed;
+        readonly IApiConnection connection;
+        bool disposed;
```
damieng: "private" is the default for class members in .net.

src/Auth0.ManagementApi/ManagementApiClient.cs
```diff
@@ -10,8 +10,8 @@ namespace Auth0.ManagementApi
     /// </summary>
     public class ManagementApiClient : IDisposable
     {
-        private readonly ApiConnection _apiConnection;
-        private bool disposed;
+        readonly IApiConnection connection;
```
damieng: Moving away from the `_` prefix for private fields.

src/Auth0.ManagementApi/ManagementApiClient.cs
```diff
@@ -206,7 +206,8 @@ protected virtual void Dispose(bool disposing)
         {
             if (!disposed && disposing)
             {
-                _apiConnection.Dispose();
+                if (connection is IDisposable disposableConnection)
+                    disposableConnection.Dispose();
```
damieng: Given that an `IApiConnection` doesn't necessarily implement `IDisposable` like `ApiConnection` does we have to check it does first before we try to dispose.

Issue #316, opened 2019-11-17
Testable - Concrete type ManagementApiClient


stephenstroud:
I want to unit test method scopes using ManagementApiClient, however, this isn't possible. Please introduce virtual so I can fake the class?

damieng:
We have a PR open to help testing at #314 that will let you wire up an alternative connection for simple mocking/stubbing.

jtbruch-fixdrepair:
@damieng looks like that PR was closed, is there a different PR that is going to make it in instead?  I am running into the same issue.

damieng:
The first one for the testability of auth is at #321 - there will be an almost identical one for management later today!

Basically you will be able to mock out the whole of the management API by just implementing an interface that has two methods - one for handling get request and another for other http requests.

jtbruch-fixdrepair:
Beautiful! Thank you so much for the quick response :) @damieng

jtbruch-fixdrepair:
@damieng thank you so much for your work on getting this testable for the group! Quick question, I saw you tagged this 7.0.0, what is your plan for releasing all this great work? :)

damieng:
We don't give firm timelines out for SDK release dates as they tend to happen when they're ready.  That said this is my top priority right now and the only remaining work is this + a tying in the symmetric id token validation.

jtbruch-fixdrepair:
Sounds good.

damieng:
Merged in, will ship in 7.0.0 - docs on how to use it will be in the migration guide.

stephenstroud:
Great!

PR #333, opened 2019-11-26
Management api connection

damieng:
Replaces `IApiConnection` and `ApiConnection` classes that are now only used by the Management Client after #321 with the equivalent `IManagementConnection` and `HttpClientManagementConnection` classes.

Also moved `FileUploadParameter` from Core to ManagementApi as it is only used there.

This PR includes a `LegacyApiConnectionBridge` class to help make it a smaller PR.  Another will follow that removes `LegacyApiConnectionBridge` and `IApiConnection`.


damieng:
Fixes #316 and #520

PR #338, opened 2019-11-26
Centralize paging code into ManagementApi.Paging

damieng:
- Only ManagementApi uses these so moved three of them from Core to ManagementApi
- Moved those + associated model & serialization classes into new ManagementApi.Paging namespace
- Documented these classes via xmldoc comments as they had either missing or incomplete docs


src/Auth0.ManagementApi/Paging/PagingInformation.cs
```diff
@@ -0,0 +1,63 @@
...
+        public int Total { get; set; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PagingInformation"/> class
+        /// with the desired arguments.
+        /// </summary>
+        /// <param name="start">Start index into the number of items.</param>
+        /// <param name="limit">Maximum number of items requested.</param>
+        /// <param name="length">Number of items actually returned.</param>
+        /// <param name="total">Total number of items available on the server.</param>
```
Widcket: Would only indicate the total number of items if `include_total` was `true` on the request.

damieng: Will add in another Pr to avoid rebasing.

PR #340, opened 2019-11-27
Improved OIDC compliance (part 3, wire up HS256) 

damieng:
This wires up the previously checked in HS256 symmetric signature verifier from #326 completing the OIDC compliance work for this SDK with the exception of Jwks caching and invalidation to follow.

Developers can chose between HS256 and RS256 depending on what their client/application is configured for. This is done on a per-request basis just like ClientId and ClientSecret.  Ideally I think the auth client should take all three pieces of information once but there have been enough breaking changes in this major.

I have made RS256 the default.  If developers are using HS256 they will need to set their requests to RS256 or they will get token validation errors.  This will be included in the migration guide.

### DIFF
--- a/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
+++ b/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
@@ -323,7 +323,7 @@ namespace Auth0.AuthenticationApi
 
                 case JwtSignatureAlgorithm.HS256:
                     if (String.IsNullOrWhiteSpace(clientSecret))
-                        throw new ArgumentException("ClientSecret must contain a value when using HS256", nameof(issuer));
+                        throw new ArgumentException("ClientSecret must contain a value when using HS256", nameof(clientSecret));
                     return SymmetricSignatureVerifier.FromClientSecret(clientSecret);
 
                 default:
--- a/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
+++ b/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
@@ -121,7 +121,7 @@ namespace Auth0.AuthenticationApi
                 body
             ).ConfigureAwait(false);
 
-            await AssertIdTokenValid(response.IdToken, request.ClientId).ConfigureAwait(false);
+            await AssertIdTokenValid(response.IdToken, request.ClientId, request.SigningAlgorithm, request.ClientSecret).ConfigureAwait(false);
 
             return response;
         }
@@ -145,7 +145,7 @@ namespace Auth0.AuthenticationApi
                 body
             ).ConfigureAwait(false);
 
-            await AssertIdTokenValid(response.IdToken, request.ClientId).ConfigureAwait(false);
+            await AssertIdTokenValid(response.IdToken, request.ClientId, request.SigningAlgorithm, request.ClientSecret).ConfigureAwait(false);
 
             return response;
         }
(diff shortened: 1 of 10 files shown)


src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
```diff
@@ -312,10 +312,32 @@ public void Dispose()
...
+            switch (algorithm)
+            {
+                case JwtSignatureAlgorithm.RS256:
+                    if (String.IsNullOrWhiteSpace(issuer))
+                        throw new ArgumentException("Issuer must contain a value when using RS256", nameof(issuer));
+                    return await AsymmetricSignatureVerifier.ForJwks(issuer).ConfigureAwait(false);
+
+                case JwtSignatureAlgorithm.HS256:
+                    if (String.IsNullOrWhiteSpace(clientSecret))
+                        throw new ArgumentException("ClientSecret must contain a value when using HS256", nameof(issuer));
```
Widcket: `issuer` should be `clientSecret`.

src/Auth0.AuthenticationApi/Models/AuthorizationCodeTokenRequest.cs
```diff
@@ -8,6 +8,6 @@ public class AuthorizationCodeTokenRequest : AuthorizationCodeRequestBase
         /// <summary>
         /// Client secret of the application.
         /// </summary>
-        public string ClientSecret { get; set; }
+        public override string ClientSecret { get; set; }
```
Widcket: Should add a comment explaining the override.

tests/Auth0.AuthenticationApi.IntegrationTests/Tokens/IdTokenValidatorIntegrationTests.cs
```diff
@@ -77,8 +77,10 @@ public async Task Passes_Token_Validation()
 
             });
 
-            var idTokenValidation = new IdTokenRequirements($"https://{GetVariable("AUTH0_AUTHENTICATION_API_URL")}/", GetVariable("AUTH0_CLIENT_ID"), TimeSpan.FromMinutes(1));
-            await idTokenValidation.AssertTokenMeetsRequirements(authenticationResponse.IdToken);
+            var issuer = $"https://{GetVariable("AUTH0_AUTHENTICATION_API_URL")}/";
+            var idTokenValidation = new IdTokenRequirements(issuer, GetVariable("AUTH0_CLIENT_ID"), TimeSpan.FromMinutes(1));
+            var verifier = await AsymmetricSignatureVerifier.ForJwks(issuer);
```
Widcket: Should add tests for symmetric as well.

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
damieng: Fixed!

src/Auth0.AuthenticationApi/Models/AuthorizationCodeTokenRequest.cs
damieng: I decided to just remove the virtual/override and have a single set of docs on the base client that indicates the optional/required is dependent on class.

tests/Auth0.AuthenticationApi.IntegrationTests/Tokens/IdTokenValidatorIntegrationTests.cs
damieng: Having some problems with symmetric integration tests so will add that in another PR to avoid more merge problems as other parts land.

PR #342, opened 2019-11-29
Remove legacy connector from ManagementApi

damieng:
Removes the legacy connection stuff leftover from #333.

As planned this will break anyone who was using `IApiConnection` in order to provide a more testable/mockable product as well as aid in future code generation efforts.

Review notes:

- Probably easier to review commit by commit
- BaseClient replaces ClientBase
- No new functionality in this, just removing the bridge and legacy interface
- Enables a few integration tests that were previously disabled (skip)
- Apologies for the size, this was actually broken out from #333 in order to keep these smaller...


src/Auth0.ManagementApi/Clients/BaseClient.cs
```diff
@@ -0,0 +1,45 @@
...
+            BaseUri = baseUri;
+        }
+
+        /// <summary>
+        /// Build the <see cref="Uri"/> for this API request given a <paramref name="resource"/> and
+        /// <paramref name="queryStrings"/>.
+        /// </summary>
+        /// <param name="resource">Name of the resource to be appended to the path.</param>
+        /// <param name="queryStrings">Query strings that should be combined with the URL.</param>
+        /// <returns>Completed absolute URi for the request.</returns>
```
joshcanhelp:
```suggestion
        /// <returns>Completed absolute URI for the request.</returns>
```

src/Auth0.ManagementApi/Clients/BaseClient.cs
```diff
@@ -0,0 +1,45 @@
...
+        protected Uri BuildUri(string resource, IDictionary<string, string> queryStrings = null)
+        {
+            return  Utils.BuildUri(BaseUri.AbsoluteUri, resource, null, queryStrings);
```
joshcanhelp:
```suggestion
            return Utils.BuildUri(BaseUri.AbsoluteUri, resource, null, queryStrings);
```

src/Auth0.ManagementApi/Clients/ClientGrantsClient.cs
```diff
@@ -1,23 +1,27 @@
...
 using System;
 using System.Collections.Generic;
+using System.Net.Http;
 using System.Threading.Tasks;
 
 namespace Auth0.ManagementApi.Clients
 {
     /// <summary>
-    /// Contains all methods to call the /client-grants endpoints.
+    /// Contains methods to acess the /client-grants endpoints.
```
joshcanhelp:
```suggestion
    /// Contains methods to access the /client-grants endpoints.
```

src/Auth0.ManagementApi/Clients/ClientGrantsClient.cs
```diff
@@ -28,7 +32,7 @@ public ClientGrantsClient(IApiConnection connection)
         /// <returns>The new <see cref="ClientGrant"/> that has been created.</returns>
         public Task<ClientGrant> CreateAsync(ClientGrantCreateRequest request)
         {
-            return Connection.PostAsync<ClientGrant>("client-grants", request, null, null, null, null, null);
+            return Connection.SendAsync<ClientGrant>(HttpMethod.Post, BuildUri("client-grants"), request);
```
joshcanhelp: **Nit:** it looks like `"client-grants"` is used enough that maybe a class-level constant would be useful here? Or maybe add to the base URI in the constructor?

src/Auth0.ManagementApi/Clients/BaseClient.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// <see cref="Uri"/> of the endpoint to use in making API calls.
+        /// </summary>
+        protected Uri BaseUri { get; }
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="BaseClient"/>.
+        /// </summary>
+        /// <param name="connection"><see cref="IManagementConnection"/> used to make all API calls.</param>
+        /// <param name="baseUri"><see cref="Uri"/> of the endpoint to use in making API calls.</param>
+        protected BaseClient(IManagementConnection connection, Uri baseUri)
```
joshcanhelp: **Question:** This is a constructor, correct? It's interesting that this class can declare its constructor as `protected` but extending class constructors are public. Is that a .NET thing or am I missing something fundamental about what's going on here?

src/Auth0.ManagementApi/Clients/ClientGrantsClient.cs
```diff
@@ -77,11 +77,7 @@ public Task<IPagedList<ClientGrant>> GetAllAsync(GetClientGrantsRequest request,
         /// <returns>The <see cref="ClientGrant"/> that has been updated.</returns>
         public Task<ClientGrant> UpdateAsync(string id, ClientGrantUpdateRequest request)
         {
-            return Connection.PatchAsync<ClientGrant>("client-grants/{id}", request, 
-                new Dictionary<string, string>
-                {
-                    {"id", id}
-                });
+            return Connection.SendAsync<ClientGrant>(new HttpMethod("PATCH"), BuildUri($"client-grants/{id}"), request);
```
joshcanhelp: Why not `HttpMethod.Patch`?

src/Auth0.ManagementApi/Clients/ClientGrantsClient.cs
joshcanhelp: **Nit (throughout):** I would not include the endpoint path in the docblocks, just the name that's used in the docs. Maybe a link to that docs page/section as well?

src/Auth0.ManagementApi/Clients/ConnectionsClient.cs
```diff
@@ -1,23 +1,27 @@
...
 using System;
 using System.Collections.Generic;
+using System.Net.Http;
 using System.Threading.Tasks;
 
 namespace Auth0.ManagementApi.Clients
 {
     /// <summary>
-    /// Contains all the methods to call the /connections endpoints.
+    /// Contains methods to acess the /connections endpoints.
```
joshcanhelp:
```suggestion
    /// Contains methods to access the /connections endpoints.
```

src/Auth0.ManagementApi/Clients/TenantSettingsClient.cs
```diff
@@ -1,21 +1,23 @@
...
+using System;
 using System.Collections.Generic;
+using System.Net.Http;
 using System.Threading.Tasks;
 
 namespace Auth0.ManagementApi.Clients
 {
     /// <summary>
-    /// Contains all the methods to call the /tenants/settings endpoints.
+    /// Contains methods to acess the /tenants/settings endpoints.
```
joshcanhelp:
```suggestion
    /// Contains methods to access the /tenants/settings endpoints.
```

src/Auth0.ManagementApi/Clients/BaseClient.cs
```diff
@@ -0,0 +1,45 @@
+﻿using Auth0.Core.Http;
+using System;
+using System.Collections.Generic;
+
+namespace Auth0.ManagementApi.Clients
+{
+    /// <summary>
+    /// Base class from which all Client classes inherit.
+    /// </summary>
+    public abstract class BaseClient
```
joshcanhelp: Good rename 👍

tests/Auth0.ManagementApi.IntegrationTests/StatsTests.cs
```diff
@@ -18,10 +18,11 @@ public async Task Test_stats_sequence()
                 // Get stats for the past 10 days
                 var dailyStats = await apiClient.Stats.GetDailyStatsAsync(DateTime.Now.AddDays(-10), DateTime.Now);
                 dailyStats.Should().NotBeNull();
+                dailyStats.Count.Should().BeGreaterOrEqualTo(1);
 
                 // Get the active users
                 var activeUsers = await apiClient.Stats.GetActiveUsersAsync();
-                activeUsers.Should().BeGreaterOrEqualTo(0);
+                activeUsers.Should().BeGreaterOrEqualTo(1); // These integration tests themselves trigger non-zero values
```
joshcanhelp: Were these not working before? If so, what changes in this PR changed this? 

Not saying you're doing anything wrong but, with these management API changes, I have definitely in the past lost sight of a functionality change because I thought that the integration tests were wonky ... turns out it was a potentially breaking change.

tests/Auth0.ManagementApi.IntegrationTests/UserBlockTests.cs
```diff
@@ -78,7 +78,7 @@ public async Task DisposeAsync()
             _apiClient.Dispose();
         }
 
-        [Fact(Skip = "Need to fix user blocks tests")]
```
joshcanhelp: 🎉

src/Auth0.ManagementApi/Clients/EmailProviderClient.cs
```diff
@@ -29,7 +31,7 @@ internal EmailProviderClient(IApiConnection connection)
         /// <returns>A <see cref="EmailProvider" /> instance containing the email provider details.</returns>
         public Task<EmailProvider> ConfigureAsync(EmailProviderConfigureRequest request)
         {
-            return Connection.PostAsync<EmailProvider>("emails/provider", request, null, null, null, null, null);
+            return Connection.SendAsync<EmailProvider>(HttpMethod.Post, BuildUri("emails/provider"), request, null);
```
joshcanhelp: Big improvement ... null ^ 5

src/Auth0.ManagementApi/Clients/EmailProviderClient.cs
```diff
@@ -38,7 +40,7 @@ public Task<EmailProvider> ConfigureAsync(EmailProviderConfigureRequest request)
         /// <returns>A <see cref="Task"/> that represents the asynchronous delete operation.</returns>
         public Task DeleteAsync()
         {
-            return Connection.DeleteAsync<object>("emails/provider", null, null);
+            return Connection.SendAsync<object>(HttpMethod.Delete, BuildUri("emails/provider"), null);
```
joshcanhelp: Does the `null` need to be explicit here? Seeing this in a few places but only just registered.

src/Auth0.ManagementApi/Clients/GuardianClient.cs
```diff
@@ -119,13 +115,12 @@ public Task<GuardianTwilioConfiguration> GetTwilioConfigurationAsync()
...
-                .PutAsync<UpdateGuardianFactorResponse>(
-                "guardian/factors/{name}",
-                new { enabled = request.IsEnabled },
-                null, null,
-                new Dictionary<string, string> { { "name", name } },
-                null, null);
+                .SendAsync<UpdateGuardianFactorResponse>(
+                HttpMethod.Put,
+                BuildUri($"guardian/factors/{name}"),
+                new { enabled = request.IsEnabled });
```
joshcanhelp: **Question:** What is `new` doing here?

src/Auth0.ManagementApi/Clients/UserBlocksClient.cs
```diff
@@ -1,21 +1,23 @@
...
+using System;
 using System.Collections.Generic;
+using System.Net.Http;
 using System.Threading.Tasks;
 
 namespace Auth0.ManagementApi.Clients
 {
     /// <summary>
-    /// Contains all the methods for the /user-blocks endpoints.
+    /// Contains methods for accessing the /user-blocks endpoints.
```
joshcanhelp:
```suggestion
    /// Contains methods to access the /user-blocks endpoints.
```

src/Auth0.ManagementApi/Clients/ClientGrantsClient.cs
damieng: Will do that in the code-generated version. Don't want to spend any more cycles on code that will be replaced.

src/Auth0.ManagementApi/Clients/ClientGrantsClient.cs
damieng: Would make code-gen more confusing and is going to be replaced by that.

src/Auth0.ManagementApi/Clients/BaseClient.cs
damieng: It basically say yes, only subclasses can call this.  It's a good practice but doesn't actually make that much difference as the class is abstract so can't be created directly anyway.

src/Auth0.ManagementApi/Clients/ClientGrantsClient.cs
damieng: Annoyingly does not exist in the .NET framework :(

src/Auth0.ManagementApi/Clients/EmailProviderClient.cs
damieng: I thought about making `null` optional on Send but I didn't want anyone to accidently wipe out user data should the method be called by a badly written or generated stub. This way you have to explicitly say that the body should be null - just in case.

src/Auth0.ManagementApi/Clients/GuardianClient.cs
damieng: It's creating a new anonymous object.  As the put only contains a single `enabled` property inside a JSON object it was deemed simpler to pass in a boolean rather than creating a Request object for it with just one property.

tests/Auth0.ManagementApi.IntegrationTests/StatsTests.cs
damieng: This test was passing... but it wasn't a good test as 0 is the default value for an enum.  So if the payload returned bad data/didn't deserialize correctly this test would have passed when it shouldn't.

We can get away with `1` as by running the tests we effectively create an active user. In theory I guess this test could be run before the number gets updated depending on the order the tests run/consistency etc. I'll keep an eye on it.

PR #347, opened 2019-12-04
JWKS caching support

damieng:
## Summary

Implements a basic form of JWKS caching.  Specifically:

1. Caches the JWKS document for 10 minutes
2. Is prepared to forcefully refresh that document every 15 seconds when keys are missing
3. Uses a static instance of `IdTokenValidator` to ensure JWKS is shared between `AuthenticationApiClient` instances

We expect to revise those numbers by the end of the beta period and also to provide some options for tweaking the caching scope.

## Refactorings to achieve this

In order to get this change in a lot of the id token validation work needed to be refactored as it was very tangled.  The changes are:

1. `IdTokenRequirements` now has an enum to specify the algorithm that is required
2. `ISignatureVerifier`s are now `SignedDecoder`s. Rather than being passed to the IdTokenValidator they are created automatically based on the `IdTokenRequirements` validator. The process now starts by a `SignedDecoder` being responsible for returning a decoded JWT only if it passes signature verification.
3. Claims validations have been split from `IdTokenValidator` to `IdTokenClaimValidator`
4. A new exception message `IdTokenValidationKeyMissingException` now exists to detect when the JWKS needs to be reconsidered for early refresh.

These also massively simplified the tests.


src/Auth0.AuthenticationApi/Tokens/IdTokenClaimValidator.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using System.IdentityModel.Tokens.Jwt;
+using System.Linq;
+using System.Security.Claims;
+
+namespace Auth0.AuthenticationApi.Tokens
+{
+    /// <summary>
+    /// Perform validation of a JWT ID token in compliance with specified <see cref="IdTokenRequirements"/>.
```
joshcanhelp:
```suggestion
    /// Perform validation of an ID token in compliance with specified <see cref="IdTokenRequirements"/>.
```

src/Auth0.AuthenticationApi/Tokens/IdTokenClaimValidator.cs
```diff
@@ -0,0 +1,112 @@
...
+    /// </summary>
+    internal static class IdTokenClaimValidator
```
joshcanhelp: This is, by far, the most clear implementation of the ID token validation stuff I've seen 💪

src/Auth0.AuthenticationApi/Tokens/IdTokenRequirements.cs
```diff
@@ -41,8 +46,9 @@ internal class IdTokenRequirements
         /// <param name="leeway">Amount of leeway in validating date and time claims to allow some clock variance
         /// between the issuer and the application.</param>
         /// <param name="maxAge">Optional maximum time since the user last authenticated.</param>
-        public IdTokenRequirements(string issuer, string audience, TimeSpan leeway, TimeSpan ? maxAge = null)
+        public IdTokenRequirements(JwtSignatureAlgorithm signatureAlgorithm, string issuer, string audience, TimeSpan leeway, TimeSpan? maxAge = null)
```
joshcanhelp: Should `leeway` be nullable here? I wouldn't require that anywhere if you can just fallback on the default.

src/Auth0.AuthenticationApi/Tokens/SignedDecoder.cs
```diff
@@ -0,0 +1,62 @@
...
+{
+    internal abstract class SignedDecoder
+    {
+        readonly JwtSignatureAlgorithm signatureAlgorithm;
+        readonly JwtSecurityTokenHandler securityTokenHandler = new JwtSecurityTokenHandler();
+        readonly TokenValidationParameters validationParameters = new TokenValidationParameters
+        {
+            ValidateAudience = false,
+            ValidateIssuer = false,
+            ValidateActor = false,
```
joshcanhelp: What's this?

src/Auth0.AuthenticationApi/Tokens/IdTokenRequirements.cs
damieng: Originally the IdTokenRequirements class/code was shared between OIDC and Auth0.NET.  As one is meant for mobile/native and the other for servers there wasn't a "good" default that applied to both so it made sense to not specify one.

There's a chance at some point we'll extract this code out into shared library both projects can use (and perhaps other people) and again I don't think there's a one-size-fits-all default for that.

src/Auth0.AuthenticationApi/Tokens/SignedDecoder.cs
damieng: We're using Microsoft's library to perform the signature validation but there's no specific "ValidateSignature" function.  Instead we have to call the general Validation function and switch all the other types of validations off.

Issue #373, opened 2020-03-06
ConfigureAwait(false) needed in Auth0.ManagementApi.HttpClientManagementConnection


cecilyth:
### Description

Calling methods on the ManagementApiClient resulted in deadlock because of missing .ConfigureAwait(false).

I was able to fix the issue by adding the ConfigureAwait(false), I'm happy to open a pull request if desired.

### Environment

Latest nuget version, WPF service stack application.

damieng:
You're right! I see two missing ones - GetAsync and SendAsync.  I can either add them or accept a PR - also would be needed on HttpClientAuthenticationConnection.

cecilyth:
> You're right! I see two missing ones - GetAsync and SendAsync. I can either add them or accept a PR - also would be needed on HttpClientAuthenticationConnection.

Great! If it's easy for you to add them quickly then that works for me! Thanks so much!

damieng:
Fixed in 7.0.2

PR #374, opened 2020-03-07
Add missing ConfigureAwait(false) to HttpClient*Connection Get/Send

damieng:
`ConfigureAwait(false)` was missing in a couple of places in `HttpClientAuthenticationConnection` and `HttpClientManagementConnection`.


damieng:
Fixes #373

PR #380, opened 2020-03-27
Capture extra error properties e.g. mfa_required

### LINKED ISSUE #376: Handling mfa_required for oauth/token endpoint
Hi there, when fetching an access token via the `oauth/token` endpoint, if you get an `mfa_required` error, there is no way to retrieve the `mfa_token` field from the `ApiException` that is thrown.

I can see this issue was solved with the Java SDK, https://github.com/auth0/auth0-java/issues/119
Would it be possible to pass back the `mfa_token` either as a property on this exception, or perhaps as part of the exception data?

Without this, we will have to hand craft our authentication requests to `oauth/token`

Thanks

damieng:
Fixes #376

### DIFF
--- a/src/Auth0.Core/Exceptions/ApiError.cs
+++ b/src/Auth0.Core/Exceptions/ApiError.cs
@@ -1,5 +1,6 @@
 ﻿using Auth0.Core.Serialization;
 using Newtonsoft.Json;
+using System.Collections.Generic;
 using System.Net.Http;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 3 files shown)


src/Auth0.Core/Exceptions/ApiError.cs
```diff
@@ -29,6 +30,12 @@ public class ApiError
         [JsonProperty("message")]
         public string Message { get; set; }
 
+        /// <summary>
+        /// Additional key/values that might be returned by the error such as `mfa_required`.
+        /// </summary>
+        [JsonProperty("extraData")]
+        public Dictionary<string, string> ExtraData { get; set; } = new Dictionary<string, string>();
```
lbalmaceda: I'm not entirely convinced of the name of the property/method "ExtraData". But I can't come up with something else right now.

damieng:
AdditionalData?

lbalmaceda:
@damieng I just checked and for classes with lots of variable properties such as a `User` instance, I use `extraInfo`. So `extraData` is fine ✅ 

In this particular scenario you're trying to cover, I expose helper functions for [checking different errors](https://github.com/auth0/Lock.Android/blob/master/lib/src/main/java/com/auth0/android/lock/LockActivity.java#L527-L528). Like "Is MFA required error ?". Then, I know I need to obtain the MFA token. For that, I have a `getValue` method that given a key it gets you the value if present. 

Your approach of exposing the whole set is also viable 👍

PR #384, opened 2020-03-31
Fix rate limit exceptions

### LINKED ISSUE #382: Unexpected Exception type
### Description
The Api returns an unexpected exception type when rate limit is reached.

### Reproduction
The following program is expected to output "rate limit reached" if given correct Url and Token but the exception is of type ```Auth0.Core.Exceptions.ErrorApiException```.

```csharp
using Auth0.Core.Exceptions;
using Auth0.ManagementApi;
using Auth0.ManagementApi.Models;
using Auth0.ManagementApi.Paging;
using System;
using System.Threading.Tasks;

namespace HammerTime
{
    class Program
    {
        public static string Token;
        public static string Url;

        async static Task Main(string[] args)
        {
            Url = args[0];
            Token = args[1];
            await HammeByApi();
        }

        private async static Task HammeByApi()
        {
            var request = new GetUsersRequest();
            var pagination = new PaginationInfo();

            var client = new ManagementApiClient(Token, new Uri(Url));
            try
            {
                while (true)
                {
                    var response = await client.Users.GetAllAsync(request, pagination);
                }
            }
            catch (RateLimitApiException)
            {
                Console.WriteLine("rate limit reached");
            }
            catch (Exception e)
            {
                Console.WriteLine($"Unexpected Exception {e.GetType().FullName} : {e.Message}");
            }
        }
    }
}

```
With the following csproj file:
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp3.1</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Auth0.ManagementApi" Version="7.0.2" />
  </ItemGroup>

</Project>
```

### Conclusion
Either return correct exception type or give some way to access returned rate limits on response.

damieng:
Fixes #382

### DIFF
--- a/src/Auth0.Core/Exceptions/ApiException.cs
+++ b/src/Auth0.Core/Exceptions/ApiException.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Net;
 using System.Net.Http;
 using System.Runtime.Serialization;
 using System.Threading.Tasks;
@@ -51,7 +50,7 @@ namespace Auth0.Core.Exceptions
         /// Create an instance of the specific exception required for this unsuccessful <see cref="HttpResponseMessage"/>.
         /// </summary>
         /// <param name="response"><see cref="HttpResponseMessage"/> to parse for the correct exception.</param>
-        /// <returns>A </returns>
+        /// <returns>An instance of a <see cref="ApiException"/> subclass containing the appropriate exception for this response.</returns>
         public static async Task<ApiException> CreateSpecificExceptionAsync(HttpResponseMessage response)
         {
             switch ((int)response.StatusCode)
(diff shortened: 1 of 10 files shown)


src/Auth0.Core/Exceptions/ApiException.cs
```diff
@@ -43,5 +46,21 @@ protected ApiException(SerializationInfo serializationInfo, StreamingContext str
             : base(serializationInfo, streamingContext)
         {
         }
+
+        /// <summary>
+        /// Create an instance of the specific exception required for this unsuccessful <see cref="HttpResponseMessage"/>.
+        /// </summary>
+        /// <param name="response"><see cref="HttpResponseMessage"/> to parse for the correct exception.</param>
+        /// <returns>A </returns>
```
jimmyjames, line 54: Looks like the `returns` doc is missing

PR #392, opened 2020-04-16
Fix import user  job

damieng:
- Fixes user import job.
- Re-enabled jobs integration tests disabled in 2017.
- Improves test coverage a little.
- Adds `send_verification_email` parameter to import users.
- Fleshes out Job class with additional possible return properties.

### DIFF
--- a/tests/Auth0.ManagementApi.IntegrationTests/Auth0.ManagementApi.IntegrationTests.csproj
+++ b/tests/Auth0.ManagementApi.IntegrationTests/Auth0.ManagementApi.IntegrationTests.csproj
@@ -7,6 +7,16 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <None Remove="user-import-test.json" />
+  </ItemGroup>
+
+  <ItemGroup>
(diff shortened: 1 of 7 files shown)


tests/Auth0.ManagementApi.IntegrationTests/Auth0.ManagementApi.IntegrationTests.csproj
```diff
@@ -6,6 +6,16 @@
     <AssemblyOriginatorKeyFile>..\..\build\Auth0NetStrongName.snk</AssemblyOriginatorKeyFile>
   </PropertyGroup>
 
+  <ItemGroup>
+    <None Remove="user-import-test.json" />
```
lbalmaceda, line 10: this is required to make the test work, right?

damieng: Yes, that tells VS to not copy the file to the output folder, the section below tells it to embed it as a resource so that the test can read it out of the exe/dll at runtime. All this only applies to the test.

PR #402, opened 2020-04-24
Preserve URL path for auth

damieng:
Right now if you pass a URL to AuthenticationApiClient the path portion is ignored.

While this is never an issue in real-world usage it does prevent some development and testing scenarios from being able to mock out the server-side.

This is a potentially breaking change if a customer today was accidentally passing a URL that incorrectly contained a path parameter but that is an unlikely scenario so not treating this as a breaking change under the "semver is an intent" philosophy.

### DIFF
--- a/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
+++ b/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
@@ -1,5 +1,6 @@
 ﻿using Auth0.AuthenticationApi.Models;
 using Auth0.AuthenticationApi.Tokens;
+using Auth0.Core.Http;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using Auth0.AuthenticationApi.Models;
 using Auth0.AuthenticationApi.Tokens;
+using Auth0.Core.Http;
```
lbalmaceda, line 3: This is for `Utils.BuildUri`?

damieng: Correct.

PR #412, opened 2020-06-23
Fix multipart boolean form casing

### LINKED ISSUE #411: Issue specifying "upsert" for ImportUsersAsync
### Description

When specifying the `upsert` parameter as `true` in the `ImportUsersAsync` method for the management API, it does not serialize correctly as Auth0 expects and instead comes out in the request as `True` (capital "T"). The Auth0 management API doesn't expect/handle this value and instead defaults `upsert` to `false`.

The only way we can bypass this at the moment seems to be not using this wrapper and going to HttpClient directly.

### Reproduction

```csharp
var jobResponse = await managementClient.Jobs.ImportUsersAsync(
                        connectionId: "con|1234",
                        fileName: "userdata.json",
                        file: stream,
                        upsert: true,
                        externalId: "123",
                        sendCompletionEmail: false);
```

### Environment

Latest Auth0.net library (7.0.8)

damieng:
Fixes #411

### DIFF
--- a/src/Auth0.ManagementApi/HttpClientManagementConnection.cs
+++ b/src/Auth0.ManagementApi/HttpClientManagementConnection.cs
@@ -156,9 +156,15 @@ namespace Auth0.ManagementApi
             if (body is Dictionary<string, object> parameters)
                 foreach (var parameter in parameters)
                     if (parameter.Value != null)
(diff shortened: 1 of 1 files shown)


lbalmaceda:
LGTM but there are some failing tests. 

```
Expected dailyStats.Count to be greater or equal to 1, but found 0.
```

damieng:
RTM.

PR #413, opened 2020-06-23
Skip unreliable stats test

damieng:
The stats tests fail if there's been little to no activity. 

Extending the daily stats out to -100 gives the daily stats test some more leeway before failure.

The active users test still fails even though sometimes it passes. The activity of the tests themselves if processed quickly enough can cause the 0 to be 1. Right now though it doesn't seem to be the case so is being skipped.


lbalmaceda:
This endpoint sounds like a candidate for moving to unit tests (no integration)

PR #617, opened 2023-05-16
Added revoke refresh token endpoint support

msmolka:
### Changes

Please describe both what is changing and why this is important. Include:

- Endpoints added
  - POST /oauth/revoke 
- Classes and methods added
  - class:  `RevokeRefreshTokenRequest`
  - method: `IAutheticationApiClient.Task<string> RevokeRefreshTokenAsync(RevokeRefreshTokenRequest request, CancellationToken cancellationToken = default)`
- A summary of usage if this is a new feature or change to a public API (this should also be added to relevant documentation once released)
  - added support for missing Revoke Refresh Token endpoint

### References

Please include relevant links supporting this change such as a:

- support ticket:  #616

### Testing

Added token test to call revoke api

- [X] This change has been tested on the latest version of the platform/language or why not

### Checklist

### DIFF
--- a/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
+++ b/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
@@ -352,6 +352,20 @@ namespace Auth0.AuthenticationApi
         }
 
         /// <inheritdoc/>
+
+        public Task<string> RevokeRefreshTokenAsync(RevokeRefreshTokenRequest request, CancellationToken cancellationToken = default)
+        {
+            if (request == null)
+                throw new ArgumentNullException(nameof(request));
+
+            return connection.SendAsync<string>(
+                HttpMethod.Post,
+                BuildUri("ouath/revoke"),
+                request,
+                cancellationToken: cancellationToken);
+        }
+
+        /// <inheritdoc/>
         public Task<SignupUserResponse> SignupUserAsync(SignupUserRequest request, CancellationToken cancellationToken = default)
         {
             if (request == null)
--- a/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
+++ b/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
@@ -369,7 +369,7 @@ namespace Auth0.AuthenticationApi
             return connection.SendAsync<string>(
                 HttpMethod.Post,
                 BuildUri("oauth/revoke"),
-                request,
+                body,
                 cancellationToken: cancellationToken);
         }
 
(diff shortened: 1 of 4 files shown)


src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
```diff
@@ -351,6 +351,20 @@ await AssertIdTokenValidIfExisting(response.IdToken, request.ClientId, request.S
...
         }
 
+        /// <inheritdoc/>
+
+        public Task<string> RevokeRefreshTokenAsync(RevokeRefreshTokenRequest request, CancellationToken cancellationToken = default)
+        {
+            if (request == null)
+                throw new ArgumentNullException(nameof(request));
+
+            return connection.SendAsync<string>(
```
frederikprijck, line 361: We need to consider the Client Assertion, like is done here: https://github.com/auth0/auth0.net/blob/7a323aee7a15eb9007ebb6db21e43bea013c3e85/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs#L140

frederikprijck (changes requested):
Thanks for the PR! There are a few things that definetly don't work yet:

- Incorrect payload sent
- Client Assertion not working at all.

I added a comment that shows how it's done. Let me know if anything is unclear.

frederikprijck (approved):
Thanks for the PR! 🚀 

FYI: I patched the PR a little bit as the endpoint does not return any data.

src/Auth0.AuthenticationApi/IAuthenticationApiClient.cs
```diff
@@ -125,6 +125,14 @@ public interface IAuthenticationApiClient : IDisposable
         /// </remarks>
         Task<AccessTokenResponse> GetTokenAsync(DeviceCodeTokenRequest request, CancellationToken cancellationToken = default);
 
+        /// <summary>
+        /// Revokes refresh token provided in request.
+        /// </summary>
+        /// <param name="request"><see cref="RevokeRefreshTokenRequest"/> containing Refresh Token and associated parameters.</param>
+        /// <param name="cancellationToken">The cancellation token to cancel operation.</param>
+        /// <returns><see cref="Task"/> representing the async operation containing either the JSON error response or the plain text success message response.</returns>
```
frederikprijck, line 133:
```suggestion
        /// <returns><see cref="Task"/> representing the async operation.</returns>
```

frederikprijck (changes requested):

frederikprijck (approved):

msmolka:
> Thanks for the PR! 🚀
> 
> FYI: I patched the PR a little bit as the endpoint does not return any data.

I think this string was ok. as I copied it from ChangePasswordAsync, when also nothing is returned, but error can be plain json. when e.g. token is empty

frederikprijck:
If the API returns a non-succesfull code, the SDK throws an exception. Even in the case when the token is empty.

ChangePassword returns a string, see the API docs: 
image

More info at https://auth0.com/docs/api/authentication#change-password

PR #647, opened 2023-08-21
Add show_as_button to ConnectionBase.cs

dylanAtWork:
### Changes

`show_as_button` is missing from Connections in auth0.net but it exists in dumps from the deploy-cli as well as in other Auth0 packages like go-auth0. This PR adds it to ConnectionBase.cs

### References

go-auth0 mentioned above:
https://github.com/auth0/go-auth0/blob/a98b898932cd024335fba6971c8a9d6853e9493c/management/connection.go#L129

I've got an application using this workaround for the time being:
```cs
    public class FixedConnectionCreateRequest : ConnectionCreateRequest
    {
        [JsonProperty("show_as_button")]
        public bool? ShowAsButton { get; set; }
    }
```

### Testing

A good test to me would probably be to create a Connection with show_as_button set, or making a GET for an existing one

### Checklist


frederikprijck:
I don't see this property being listed in the API docs: 

- https://auth0.com/docs/api/management/v2/connections/get-connections
- https://auth0.com/docs/api/management/v2/connections/post-connections
- https://auth0.com/docs/api/management/v2/connections/patch-connections-by-id

I have reached out internally to understand why that property isn't documented. Typically it means that it should be used publicly, and should not be part of our SDK(s).

However, I will get back to this once I have more information.

dylanAtWork:
That's odd, it's something new with Universal Login I think so maybe it was just missed?

it corresponds with this in the Admin UI
image

frederikprijck (dismissed):

frederikprijck (approved):

frederikprijck:
Thanks for the PR, sorry it took so long but I needed to get some information internally.

We will need to circle back to update XML docs here, but will merge.

PR #678, opened 2023-11-24
Added support for CheckpointPagingInformation on Connections.GetAllAsync

LaurentLardenois:
Added support for CheckpointPagingInformation on Connections.GetAllAsync to support the management api feature according to the docs (https://auth0.com/docs/api/management/v2/connections/get-connections). Added a base function with overloads and extra documentation for clarification. These changes allow the retrieval of all connections on tenants with >1000 connections. To test just try to receive connections on a tenant with all three overloads.

### DIFF
--- a/src/Auth0.ManagementApi/Clients/ConnectionsClient.cs
+++ b/src/Auth0.ManagementApi/Clients/ConnectionsClient.cs
@@ -114,13 +114,16 @@ namespace Auth0.ManagementApi.Clients
         {
             if (request == null)
                 throw new ArgumentNullException(nameof(request));
+            
+            if (pagination == null)
+                throw new ArgumentNullException(nameof(pagination));
 
             var queryStrings = new Dictionary<string, string>();
             queryStrings["page"] = pagination.PageNo.ToString();
             queryStrings["per_page"] = pagination.PerPage.ToString();
             queryStrings["include_totals"] = pagination.IncludeTotals.ToString().ToLower();
 
-            return GetAllAsync(request, new Dictionary<string, string>(), cancellationToken);
+            return GetAllAsync(request, queryStrings, cancellationToken);
         }
 
         /// <summary>
@@ -134,6 +137,9 @@ namespace Auth0.ManagementApi.Clients
         {
             if (request == null)
                 throw new ArgumentNullException(nameof(request));
+            
+            if (pagination == null)
+                throw new ArgumentNullException(nameof(pagination));
 
             var queryStrings = new Dictionary<string, string>();
             queryStrings["take"] = pagination.Take.ToString();
@@ -142,7 +148,7 @@ namespace Auth0.ManagementApi.Clients
                 queryStrings["from"] = pagination.From.ToString();
             }
 
-            return GetAllAsync(request, new Dictionary<string, string>(), cancellationToken);
+            return GetAllAsync(request, queryStrings, cancellationToken);
         }
 
         /// <summary>
--- a/src/Auth0.ManagementApi/Clients/ConnectionsClient.cs
+++ b/src/Auth0.ManagementApi/Clients/ConnectionsClient.cs
@@ -86,28 +86,77 @@ namespace Auth0.ManagementApi.Clients
         /// <summary>
         /// Retrieves every connection matching the specified strategy. All connections are retrieved if no strategy is being specified. Accepts a list of fields to include or exclude in the resulting list of connection objects.
         /// </summary>
+        /// <remarks>
+        /// Can only retrieve a maximum of 1000 connections, if you need to retrieve more use the overload with <see cref="CheckpointPagingInformation"/>
+        /// </remarks>
         /// <param name="request">Specifies criteria to use when querying connections.</param>
-        /// <param name="pagination">Specifies pagination info to use when requesting paged results.</param>
         /// <param name="cancellationToken">The cancellation token to cancel operation.</param>
         /// <returns>An <see cref="IPagedList{Connection}"/> containing the list of connections.</returns>
-        public Task<IPagedList<Connection>> GetAllAsync(GetConnectionsRequest request, PaginationInfo pagination = null, CancellationToken cancellationToken = default)
+        private Task<IPagedList<Connection>> GetAllAsync(GetConnectionsRequest request, CancellationToken cancellationToken = default)
         {
             if (request == null)
                 throw new ArgumentNullException(nameof(request));
 
-            var queryStrings = new Dictionary<string, string>
-            {
-                {"fields", request.Fields},
-                {"include_fields", request.IncludeFields?.ToString().ToLower()},
-                {"name", request.Name},
-            };
+            return GetAllAsync(request, new Dictionary<string, string>(), cancellationToken);
+        }
 
-            if (pagination != null)
+        /// <summary>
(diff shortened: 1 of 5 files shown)


src/Auth0.ManagementApi/Clients/ConnectionsClient.cs
```diff
@@ -86,29 +86,84 @@ public Task<Connection> GetAsync(string id, string fields = null, bool includeFi
...
         /// <param name="cancellationToken">The cancellation token to cancel operation.</param>
         /// <returns>An <see cref="IPagedList{Connection}"/> containing the list of connections.</returns>
-        public Task<IPagedList<Connection>> GetAllAsync(GetConnectionsRequest request, PaginationInfo pagination = null, CancellationToken cancellationToken = default)
+        public Task<IPagedList<Connection>> GetAllAsync(GetConnectionsRequest request, PaginationInfo pagination, CancellationToken cancellationToken = default)
         {
             if (request == null)
                 throw new ArgumentNullException(nameof(request));
+            
+            if (pagination == null)
+                throw new ArgumentNullException(nameof(pagination));
```
frederikprijck, line 119: This is a breaking change, anyone that used to do `GetAllAsync(new GetConnectionsRequest(), null)`, will now get an error.

We shouldn't break existing behavior.

frederikprijck (changes requested):
Thanks for opening the PR. I had a first quick look and this branch doesn't even compile.

Happy to give this another review when this is ready and properly compiles.

src/Auth0.ManagementApi/Clients/ConnectionsClient.cs
LaurentLardenois: My last commit should have resolved this issue

src/Auth0.ManagementApi/Clients/ConnectionsClient.cs
```diff
@@ -104,9 +104,18 @@ public Task<IPagedList<Connection>> GetAllAsync(GetConnectionsRequest request, P
...
+                    case PaginationInfo offsetPagination:
+                        queryStrings["page"] = offsetPagination.PageNo.ToString();
+                        queryStrings["per_page"] = offsetPagination.PerPage.ToString();
+                        queryStrings["include_totals"] = offsetPagination.IncludeTotals.ToString().ToLower();
+                        break;
+                    case CheckpointPaginationInfo checkpointPagination:
+                        queryStrings["from"] = checkpointPagination.From;
+                        queryStrings["take"] = checkpointPagination.Take.ToString();
+                        break;
+                }
```
frederikprijck, line 118: We shouldnt need this. All we had to avoid is introducing the breaking change. We have no need for BasePaginationInfo or such a switch case.

Let's just add a new method for the checkpoint pagination where checkpoint pagination is required.

frederikprijck (changes requested):

src/Auth0.ManagementApi/Clients/ConnectionsClient.cs
LaurentLardenois: What is the software design choice to use an overload, but also use one with defaults? There seems to be a convolution of different architecture choices. I would argue that with the base class it stays more in line with a singular way of working with defaults, which seems to be the current 'standard'. For instance that link you provide were the situation is 'the other way around' already is a different way of working since the other overload is the default. I know that switching everything to the baseclass variant with these changes also changes a bit, but in my opinion the base class stays more in line with the way of working with defaults. Let me know your thoughts!

LaurentLardenois: Also, calling Connection.GetAllAsync(request, null, cancellationToken) will result in an ambiguous overload error. Which is also the case with the ClientGrantsClient.GetAllOrganizationsAsync call.

LaurentLardenois: image
See here the example, so that would mean that with the change you suggested, it would still be a breaking change for everybody that explicitly sets the value of pagination to null.

frederikprijck: Why could we not stick to what you had initially, but avoid throwing the error if the paging was null?

LaurentLardenois: Because that would still result in an ambiguous invocation (unless you want to upgrade the whole SDK to a framework version that supports nullable types and make only one overload nullable that way, and I'm not even sure that would work). The issue is that the sdk has not been designed to work with overloads properly, and therefore we should work towards a way that makes these ambiguous calls impossible and forces a strict way of working. For now the best way to have it not breaking and allow for future expansion is in my opinion to work with the base class with default null and type switched branching for implementation. I'm not really a big fan of default values for this exact reason. These issues would never have surfaced if from the beginning there were stricter overloads that never accepted null, because than this would never break the backwards compatibility. But for now I think this latest version with the base class (I hate empty abstract base classes and interfaces so believe me I chose the lesser of the evils here) is the best solution we have.

frederikprijck: Having looked into this, let's just revert the last two commits (we need to revert https://github.com/auth0/auth0.net/pull/678/commits/f1e03447dd19c47f5bb55c77366c2a695481b024 as well, because that is general guidance and not related to this change) and ensure things compile (it does not compile when you revert those last two commits because you use `CheckpointPagingInformation` instead of `CheckpointPaginationInfo`. 

Making that change will result in a compilation error for anyone that is passing null explicitly and that should be acceptable. We try and stick to semver wherever possible, but compile time breaking changes are arguably acceptable in certain situations.

> The issue is that the sdk has not been designed to work with overloads properly

I believe our overloads are not bad, but it's not possible to introduce this in a non breaking way so there is definitely an opportunity to improve. Just not in this PR.

We will also need to patch the interface so that it matches the implementation and add tests.

frederikprijck: And to be clear, I appreciate the work you put in this and the idea you bring to the table about the pagination. I just think we should not just randomly change how we handle pagination overloads in a PR like this, but properly consider how we want to implement this going forward.

LaurentLardenois: Since reverting this current PR to the suggested commit would introduce breaking changes for people, I think changing this PR into the pagination refactor everywhere is a good idea. The question than is, what is the preference? With the base class it would not lead to a breaking change anywhere, even were people are currently using the overloads, however if we turn everything into overloads it will introduce breaking changes, and well I would argue this is the better solution from a software architecture standpoint, deviating only with this would make the code base more inconsistent in my opinion.
So I leave it up to you to choose between refactoring all pagination functionality with either non nullable overloads or a base class for the pagination. Note that in case of the base class I will have to look on an endpoint to endpoint basis if checkpoint pagination is allowed and make sure to put it in the documentation of the function and throw not supported exceptions if checkpoint pagination info is passed to the function, or you can choose to leave the strict pagination type intact in places that do not support checkpoint pagination yet.
Let me know your thoughts!

frederikprijck:
Given the discussion we had, I am going to close this PR. Feel free to open an issue to discuss the pagination changes if you want.

Regardless, I appreciate the effort you put in this, thanks!

Issue #882, opened 2025-09-24
Native to Web SSO Support


mfolker-sage:
### Checklist

- [x] I have looked into the [Readme](https://github.com/auth0/auth0.net#readme) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://auth0.github.io/auth0.net/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/auth0.net/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

Currently the management api SDK does not support native to web sso. Therefore tooling written to manage clients using that SDK cannot be used to configure native to web sso.

https://auth0.com/docs/authenticate/single-sign-on/native-to-web

https://auth0.com/docs/authenticate/single-sign-on/native-to-web/configure-implement-native-to-web

### Describe the ideal solution

Add support for the following properties which can be included on GET, POST, PATCH requests for clients 

{
  "session_transfer": {
    "can_create_session_transfer_token": false,
    "allowed_authentication_methods": ["cookie", "query"],
    "enforce_device_binding": "ip", // also "none" or "asn",
    "allow_refresh_token": false,
    "enforce_cascade_revocation": true,
    "enforce_online_refresh_tokens": true
  }
}

PR #885, opened 2025-09-26
Adds support to manage Native to SSO configuration on clients

kailash-b:
### ✏️ Changes

This change adds support for managing Native to SSO (Single Sign-On) configuration on Auth0 clients through the Management API.

**Classes and methods added:**
- New `SessionTransfer` model class with properties for configuring native-to-web SSO behavior
- Added `SessionTransfer` property to `ClientBase` class

**Properties added to SessionTransfer model:**
- `CanCreateSessionTransferToken` - Controls whether an app can issue session tokens through Token Exchange
- `AllowedAuthenticationMethods` - Specifies which methods can create sessions from session tokens
- `EnforceDeviceBinding` - Configures device binding security (by IP, ASN, or none)
- `AllowRefreshToken` - Controls refresh token issuance during session transfer authentication
- `EnforceOnlineRefreshTokens` - Ties refresh tokens to session lifetime
- `EnforceCascadeRevocation` - Enables cascade revocation of dependent tokens

**Usage summary:**
This feature allows developers to configure Native to Web SSO settings when creating or updating Auth0 clients. The `SessionTransfer` property can be set on client creation/update requests to control how native applications can transfer authentication sessions to web applications.

###  🔗 References

- [SDK-6872](https://auth0team.atlassian.net/browse/SDK-6872)
- #882 

### 🎯 Testing

This change has been tested through integration tests that verify:
- Creating clients with session transfer configuration
- Updating existing clients with session transfer settings  
- Reading back session transfer configuration from created/updated clients
- All session transfer properties are properly serialized/deserialized

**Testing coverage:**
- Integration tests added in `ClientTests.cs` covering full CRUD operations with session transfer configuration
- Tests verify all boolean, string, and array properties are handled correctly
- Tests cover both enabled and disabled states for all configuration options

- [x] This change has been tested on the latest version of the platform/language

### ✅ Checklist

- [x] All existing and new tests complete

[SDK-6872]: https://auth0team.atlassian.net/browse/SDK-6872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Auth0.ManagementApi/Models/Client/ClientBase.cs
+++ b/src/Auth0.ManagementApi/Models/Client/ClientBase.cs
@@ -223,7 +223,9 @@ public abstract class ClientBase
     [JsonProperty("token_quota")]
     public TokenQuota TokenQuota { get; set; }
     
-    ///  <inheritdoc cref="Auth0.ManagementApi.Models.SessionTransfer"/>
+    /// <summary>
+    /// Native to Web SSO configuration.
+    /// </summary>
     [JsonProperty("session_transfer")]
     public SessionTransfer? SessionTransfer { get; set; }
 }
--- a/src/Auth0.ManagementApi/Models/Client/DeviceBindingType.cs
+++ b/src/Auth0.ManagementApi/Models/Client/DeviceBindingType.cs
@@ -1,0 +1,32 @@
+using System.Runtime.Serialization;
+
+namespace Auth0.ManagementApi.Models;
+
+/// <summary>
+/// Indicates whether device binding security should be enforced for the app.
+/// If set to 'ip', the app will enforce device binding by IP, meaning that consumption of session_token must be
+/// done from the same IP of the issuer. Likewise, if set to 'asn', device binding is enforced by ASN, meaning
+/// consumption of session_token must be done from the same ASN as the issuer.
+/// If set to 'null', device binding is not enforced.
+/// </summary>
+public enum DeviceBindingType
+{
+    /// <summary>
+    /// Ip
+    /// </summary>
+    [EnumMember(Value = "ip")]
+    Ip,
+    
+    /// <summary>
+    /// Asn
+    /// </summary>
+    [EnumMember(Value = "asn")]
+    Asn,
+    
+    /// <summary>
+    /// None
+    /// </summary>
+    [EnumMember(Value = "none")]
+    None,
+    
+}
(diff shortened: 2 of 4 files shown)


kailash-b:
@claude

kailash-b:
@claude

mfolker-sage:
Also, I would appreciate being notified when this is merged and pushed to the public nuget server if possible.

kailash-b:
Hi @mfolker-sage 👋 

> Can I suggest using an enum for AllowedAuthenticationMethods, i.e.
> 
> ```
> [JsonProperty("allowed_authentication_methods", ItemConverterType = typeof(StringEnumConverter))]
> public SessionTransferAuthenticationMethod[]? AllowedAuthenticationMethods { get; set; }
> ```
> 
> ///
> 
> 
> /// Allowed authentication methods for session transfer.
> ///
> 
> public enum SessionTransferAuthenticationMethod
> {
> [EnumMember(Value = "cookie")]
> Cookie,
> [EnumMember(Value = "query")]
> Query
> }

This would have ideally been the approach if the allowed values were clearly mentioned on the [auth0/docs](https://auth0.com/docs/api/management/v2/clients/patch-clients-by-id). Since we do not know what the allowed values are, we let it be as is. If in the future these values are mentioned in the docs, then we will also update the SDK accordingly. 




> Also, I would appreciate being notified when this is merged and pushed to the public nuget server if possible.

Sure! We will keep the original issue updated as soon as this fix is released and is available for use.

kailash-b:
> > This would have ideally been the approach if the allowed values were clearly mentioned on the [auth0/docs](https://auth0.com/docs/api/management/v2/clients/patch-clients-by-id). Since we do not know what the allowed values are, we let it be as is. If in the future these values are mentioned in the docs, then we will also update the SDK accordingly.
> 
> image
> https://auth0.com/docs/authenticate/single-sign-on/native-to-web/configure-implement-native-to-web
> 
> If it helps, I think it is pretty clear in the documentation above.

That is correct. However, the SDK implementation is bound to the API's [schema](https://auth0.com/docs/api/management/v2/clients/patch-clients-by-id). Since the schema does not restrict the users from sending any string, we will not be able to add that constraint in the SDK for now. 

In the meantime, we will initiate an internal discussion on whether this can be made explicit in the schema as well. That will make it clear for everyone. Once that reaches a conclusion, we will update the type to an `enum`

PR #1046, opened 2026-07-08
feat: Phasing out Newtonsoft.Json in favour of System.Text.Json

kailash-b:
### Changes

This PR phases out **Newtonsoft.Json** in favour of **System.Text.Json (STJ)** across
`Auth0.Core` and `Auth0.AuthenticationApi`, bringing them in line with `Auth0.ManagementApi`
(already on STJ) and removing the Newtonsoft dependency from the serialization path. The
migration is designed for **zero-to-minimum breaking changes** for consumers.

**Auth0.Core**
- Added `Auth0JsonSerializerOptions.Default` — a shared internal options singleton
  (`PropertyNameCaseInsensitive = true`, `DefaultIgnoreCondition = WhenWritingNull`) that
  reproduces Newtonsoft's case-insensitive matching and `NullValueHandling.Ignore`.
- Rewrote `ApiErrorConverter` as an STJ `JsonConverter<ApiError>`.
- Migrated `StringOrObjectAsStringConverter` and `StringOrStringArrayJsonConverter` to
  `JsonConverter<T>`. These converters are `internal`, so the base-type change is not part
  of the public contract.

**Auth0.AuthenticationApi**
- Migrated all model types and `HttpClientAuthenticationConnection` to STJ
  (`[JsonProperty]` → `[JsonPropertyName]`, Newtonsoft `[JsonIgnore]` → STJ `[JsonIgnore]`,
  serialize/deserialize via the shared options). Serialization uses `body.GetType()` so
  polymorphic/derived properties are still emitted.

**Public API impact (deprecation, not removal)**
- `UserInfo.AdditionalClaims` (`IDictionary<string, JToken>`) is **kept** but marked
  `[Obsolete]` and `[JsonIgnore]`. It is now a computed shim that reads from a new
  `AdditionalClaimsJson` property. Existing consumer code that reads `AdditionalClaims`
  continues to compile and still returns `JToken`s.
- Added `UserInfo.AdditionalClaimsJson` (`IDictionary<string, JsonElement>`) as the real
  `[JsonExtensionData]` sink (additive; defaults to an empty dictionary so the obsolete
  accessor is null-safe).
- No other public property was renamed, retyped, or removed - all remaining model changes
  are attribute-only.

**Intentional behavioural delta**
- `StringOrObjectAsStringConverter` now emits **compact** JSON where Newtonsoft emitted
  pretty-printed output (e.g. `{"innerValue":"test"}` instead of the indented form). This is
  asserted explicitly in the tests.

### Testing

### Checklist


nandan-bhat (commented):
Left a few inline comments on the Newtonsoft to System.Text.Json migration. Overall it looks solid and the converter/UserInfo test coverage is good. My main worry is the `SignupUserResponse` private field handling, plus a couple of spots where the behaviour quietly changes between the two libraries.

src/Auth0.AuthenticationApi/Models/SignupUserResponse.cs
```diff
@@ -29,68 +29,72 @@ public class SignupUserResponse
...
     public string Id
     {
         get { return _id ?? id ?? user_id; }
         set { _id = value; }
     }
 
-    [JsonProperty("_id")]
+    [JsonInclude]
+    [JsonPropertyName("_id")]
     private string _id;        // Standard connection
```
nandan-bhat: These backing fields are `private` and rely on `[JsonInclude]` to get populated. `System.Text.Json` does not treat `[JsonInclude]` on private fields the same way Newtonsoft did. Depending on the runtime it either ignores them or throws when it builds the contract. If they get ignored, `Id` will always be `null` because none of `_id`, `id` or `user_id` ever get set.

I could not find any test that deserializes `SignupUserResponse`, so this would slip through. Can we add a test that feeds `_id`, `id` and `user_id` payloads separately and checks `Id` resolves in each case? If the attribute does not work on `net462` / `netstandard2.0`, we may need the custom `TypeInfoResolver` approach that `ManagementApi` already uses, or make these members non private.

src/Auth0.AuthenticationApi/Models/SignupUserResponse.cs
```diff
@@ -29,68 +29,72 @@ public class SignupUserResponse
...
-    [JsonProperty("picture")]
+    [JsonPropertyName("picture")]
     public Uri Picture { get; set; }
 
     /// <summary>
     /// Metadata the user has read/write access to. 
     /// </summary>
-    [JsonProperty("user_metadata")]
+    [JsonPropertyName("user_metadata")]
     public dynamic UserMetadata { get; set; }
```
nandan-bhat, line 82: Heads up: `UserMetadata` has no converter, so under `System.Text.Json` this deserializes to a `JsonElement` instead of the `JObject` Newtonsoft used to give. Any consumer doing `dynamic md = response.UserMetadata; md.someField` will now throw `RuntimeBinderException`. We kept the old shape for `UserInfo.AdditionalClaims` through the obsolete shim, but this one changes silently. At the very least we should list it as a breaking change.

src/Auth0.AuthenticationApi/Models/PasswordlessSmsResponse.cs
```diff
@@ -10,18 +10,18 @@ public class PasswordlessSmsResponse
...
-    [JsonProperty("phone_number")]
+    [JsonPropertyName("phone_number")]
     public string PhoneNumber { get; set; }
 
     /// <summary>
     /// Language the message sent was written in.
     /// </summary>
-    [JsonProperty("request_language")]
+    [JsonPropertyName("request_language")]
     public object RequestLanguage { get; set; }
```
nandan-bhat, line 31: Same story as `UserMetadata`. This was a `JValue` under Newtonsoft and becomes a `JsonElement` now, so a cast like `(string)response.RequestLanguage` will throw `InvalidCastException`. Worth calling out as a behaviour change, or exposing it as `JsonElement` on purpose.

src/Auth0.Core/Serialization/ApiErrorConverter.cs
```diff
@@ -1,59 +1,63 @@
...
-    }
-
-    public override object ReadJson(JsonReader reader, Type objectType, object? existingValue, JsonSerializer serializer)
-    {
-        var instance = Activator.CreateInstance(objectType);
-        var props = objectType.GetTypeInfo().DeclaredProperties.Where(p => p.CanWrite).ToList();
-        var extraDataProp = props.FirstOrDefault(p => p.PropertyType == typeof(Dictionary<string, string>));
-
-        foreach (var jp in JObject.Load(reader).Properties())
+        foreach (var property in doc.RootElement.EnumerateObject())
```
nandan-bhat, line 24: `EnumerateObject()` throws `InvalidOperationException` when the content is valid JSON but not an object (for example `123`, a plain string, or an array). That is not a `JsonException`, so the try/catch fallback in `ApiError.Parse` will not catch it and it escapes.

The old Newtonsoft path threw a `JsonException` here, so the fallback used to kick in and return the raw content. Can we add a guard like `if (doc.RootElement.ValueKind != JsonValueKind.Object)` and return early, plus a test for something like `ApiError.Parse("123")`?

src/Auth0.AuthenticationApi/FlexibleDateTimeConverter.cs
```diff
@@ -1,41 +1,36 @@
...
     {
-        throw new NotImplementedException();
+        switch (reader.TokenType)
+        {
+            case JsonTokenType.Null:
+                return null;
+            case JsonTokenType.Number:
+                return Add(Epoch, TimeSpan.FromSeconds(reader.GetInt64()));
+            case JsonTokenType.String:
+                return reader.GetDateTime();
```
nandan-bhat: `reader.GetDateTime()` only accepts `ISO 8601` style strings and throws `FormatException` on anything else. The old converter was lenient and just returned the raw value. So for any non ISO date string this now throws instead of degrading. Auth0 normally sends epoch or ISO here so the real risk is low, but a `DateTime.TryParse` fallback (or returning `null`) would keep the old leniency.

src/Auth0.Core/Auth0.Core.csproj
```diff
@@ -16,6 +16,7 @@
   <!-- Package Dependencies -->
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="13.0.4" />
+    <PackageReference Include="System.Text.Json" Version="8.0.5" />
```
nandan-bhat: Small consistency thing: `Core` pins `System.Text.Json` to `8.0.5` while `AuthenticationApi` and `ManagementApi` are on `9.0.9`. NuGet will unify to `9.0.9` at runtime so it builds fine, but the different floor here is easy to trip over later, and `8.0.x` is a separate servicing line for security patches. Can we bump this to `9.0.9` to match?

src/Auth0.Core/Serialization/StringOrStringArrayJsonConverter.cs
```diff
@@ -1,39 +1,33 @@
...
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
+using System.Collections.Generic;
+using System.Text.Json;
+using System.Text.Json.Serialization;
 
 namespace Auth0.Core.Serialization;
 
-internal class StringOrStringArrayJsonConverter : JsonConverter
+internal class StringOrStringArrayJsonConverter : JsonConverter<object?>
```
nandan-bhat: Nit: I could not find any `[JsonConverter(typeof(StringOrStringArrayJsonConverter))]` usage in `src`, so this looks unused. If it is dead code we could drop it along with its tests instead of porting it over. Feel free to ignore if it is kept on purpose for something planned.

src/Auth0.AuthenticationApi/Models/SignupUserResponse.cs
kailash-b: This is fixed with a custom serialiser in [commit](7af98df7927c30465e90fd5630de8e9a8ee89210)

src/Auth0.AuthenticationApi/Models/SignupUserResponse.cs
kailash-b: Documented the [behaviour](7af98df7927c30465e90fd5630de8e9a8ee89210). We will also call this out as a breaking change on release. 
The field is `dynamic` which already means there is no guarantee on the type (structure). However, since we are knowingly breaking this type, we can ensure we call this out explicitly.

src/Auth0.AuthenticationApi/Models/PasswordlessSmsResponse.cs
kailash-b: Same as above.

src/Auth0.Core/Serialization/ApiErrorConverter.cs
kailash-b: Addressed!

src/Auth0.AuthenticationApi/FlexibleDateTimeConverter.cs
kailash-b: Addressed

src/Auth0.Core/Auth0.Core.csproj
kailash-b: Bumped to `9.0.9`

src/Auth0.Core/Serialization/StringOrStringArrayJsonConverter.cs
kailash-b: Deleted dead code

nandan-bhat (approved):

PR #1056, opened 2026-07-15
feat: Adds MRRT support

kailash-b:
### Changes

Adds **Multi-Resource Refresh Token (MRRT)** support to the `Auth0.AuthenticationApi` package.

MRRT lets a single refresh token be exchanged for access tokens targeting different APIs (`audience`) and/or broader scopes, without a full re-authentication. At the protocol level this is already a standard `POST /oauth/token` refresh-token grant — `RefreshTokenRequest` already carries `Audience` and `Scope`, and `GetTokenAsync(RefreshTokenRequest)` already sends them. The gap this PR closes is on the **response** side: surfacing the actually-granted scopes so callers can detect under-granting.

**Classes/methods changed:**
- `AccessTokenResponse` — added `Scope` property (`[JsonPropertyName("scope")]`). 
The token endpoint returns the granted scope, which may be narrower than requested. Promoting it to the base response makes it available to every flow that returns `AccessTokenResponse`.
 - ClientInitiatedBackchannelAuthorizationTokenResponse — removed its duplicate Scope property; it now inherits from the base AccessTokenResponse. This is source-compatible but binary-breaking: the property
name, wire name (scope), type, and serialization are unchanged, so any code recompiled against this version behaves identically. However, an assembly compiled against a previous version that reads
`ClientInitiatedBackchannelAuthorizationTokenResponse.Scope` holds a member reference to the derived type; because the property now lives on the base, that reference will throw `MissingMethodException` at runtime
until the consuming assembly is recompiled.
- `AuthenticationApiClient.GetTokenAsync(RefreshTokenRequest)` — changed `AssertIdTokenValid` → `AssertIdTokenValidIfExisting`. A refresh-token grant only returns an `id_token` when `openid` scope is requested, which MRRT audience-targeted refreshes frequently omit; the previous call threw `"ID token is required but missing."` on an absent token. This aligns the refresh flow with the client-credentials, passwordless, and device-code flows, which already validate the ID token only when present. Backward-compatible: refresh responses that *do* carry an `id_token` are still validated.

**Usage** (also added to `Examples.md`, §5):
```csharp
var requestedScope = "read:data write:data";

var tokenResponse = await authClient.GetTokenAsync(new RefreshTokenRequest
{
    ClientId = _clientId,
    ClientSecret = _clientSecret,
    RefreshToken = _refreshToken,
    Audience = "https://my-api.example.com", // target a different API (optional)
    Scope = requestedScope                    // request broader scopes (optional)
});

// AccessTokenResponse.Scope reflects what was actually granted — may be narrower.
if (tokenResponse.Scope != requestedScope)
{
    // handle insufficient-scope grant
}
```

### Testing

### Checklist

### DIFF
--- a/Examples.md
+++ b/Examples.md
@@ -17,6 +17,7 @@
 - [4. Rich Authorization Requests (RAR)](#4-rich-authorization-requests-rar)
   - [4.1. Pushed Authorization Request (PAR) with authorization details](#41-pushed-authorization-request-par-with-authorization-details)
   - [4.2. Client Initiated Backchannel Authorization (CIBA) with authorization details](#42-client-initiated-backchannel-authorization-ciba-with-authorization-details)
+- [5. Multi-Resource Refresh Token (MRRT)](#5-multi-resource-refresh-token-mrrt)
 
 ## 1. Client Initialization
 
@@ -257,6 +258,41 @@ if (tokenResponse.AuthorizationDetails != null)
     {
         Console.WriteLine(detail.Type);
     }
+}
+```
+
+[Go to Top](#)
+
+## 5. Multi-Resource Refresh Token (MRRT)
+
+A Multi-Resource Refresh Token lets you exchange a single refresh token for access
+tokens targeting different APIs (`Audience`) and/or broader scopes, without a full
+re-authentication. At the protocol level this is a standard refresh-token grant with
+`Audience` and/or `Scope` set.
+
+The token endpoint returns the scopes that were **actually granted** in
+`AccessTokenResponse.Scope` (RFC 6749 §5.1), which may be narrower than what you
+requested. Compare the two to detect an insufficient-scope grant.
+
+```csharp
+var requestedScope = "read:data write:data";
+
+var tokenResponse = await authClient.GetTokenAsync(new RefreshTokenRequest
+{
+    ClientId = _clientId,
+    ClientSecret = _clientSecret,
+    RefreshToken = _refreshToken,
+    Audience = "https://my-api.example.com", // target a different API (optional)
+    Scope = requestedScope                    // request broader scopes (optional)
+});
+
+Console.WriteLine($"Access Token : {tokenResponse.AccessToken}");
+Console.WriteLine($"Granted Scope: {tokenResponse.Scope}");
+
+// The server may grant fewer scopes than requested — check before relying on them.
+if (tokenResponse.Scope != requestedScope)
+{
+    Console.WriteLine("Warning: not all requested scopes were granted.");
 }
 ```
 
--- a/Examples.md
+++ b/Examples.md
@@ -289,10 +289,14 @@ var tokenResponse = await authClient.GetTokenAsync(new RefreshTokenRequest
 Console.WriteLine($"Access Token : {tokenResponse.AccessToken}");
 Console.WriteLine($"Granted Scope: {tokenResponse.Scope}");
 
-// The server may grant fewer scopes than requested — check before relying on them.
-if (tokenResponse.Scope != requestedScope)
+// The server may grant fewer scopes than requested, and returns them in no guaranteed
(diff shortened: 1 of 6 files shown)


Examples.md
```diff
@@ -262,6 +263,41 @@ if (tokenResponse.AuthorizationDetails != null)
...
+    RefreshToken = _refreshToken,
+    Audience = "https://my-api.example.com", // target a different API (optional)
+    Scope = requestedScope                    // request broader scopes (optional)
+});
+
+Console.WriteLine($"Access Token : {tokenResponse.AccessToken}");
+Console.WriteLine($"Granted Scope: {tokenResponse.Scope}");
+
+// The server may grant fewer scopes than requested — check before relying on them.
+if (tokenResponse.Scope != requestedScope)
```
nandan-bhat: This comparison is not reliable. The `scope` value is a space separated list and the server does not guarantee any ordering. If the server grants everything but returns the scopes in a different order (like `write:data read:data`), this check will still report a warning even though nothing is actually missing.

Can we compare them as sets instead? Something like splitting both strings on spaces and checking that every requested scope is present in the granted ones. Since this example gets copied into real apps, I would rather we show the correct pattern here.

src/Auth0.AuthenticationApi/Models/AccessTokenResponse.cs
```diff
@@ -25,6 +25,17 @@ public class AccessTokenResponse : TokenBase
...
+    /// <summary>
+    /// The scopes that were actually granted for this token.
+    /// </summary>
+    /// <remarks>
+    /// The authorization server returns the granted scope, which
+    /// may be narrower than the scope that was requested. Compare this against the
+    /// requested scope to detect an insufficient-scope grant.
+    /// </remarks>
+    [JsonPropertyName("scope")]
+    public string Scope { get; set; }
```
nandan-bhat, line 38: Moving `Scope` up to the base class is fine for source code, but it is a binary breaking change. Any app that was compiled against the older version and reads `ClientInitiatedBackchannelAuthorizationTokenResponse.Scope` will throw a `MissingMethodException` at runtime until it is recompiled.

In practice most people recompile with the new package, so the impact is small, but the PR description says this is non breaking. Can we update the description or the release notes to call this out?

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
```diff
@@ -220,7 +220,7 @@ public async Task<AccessTokenResponse> GetTokenAsync(RefreshTokenRequest request
             cancellationToken: cancellationToken
         ).ConfigureAwait(false);
 
-        await AssertIdTokenValid(response.IdToken, request.ClientId, request.SigningAlgorithm, request.ClientSecret, request.Organization, request.Nonce).ConfigureAwait(false);
+        await AssertIdTokenValidIfExisting(response.IdToken, request.ClientId, request.SigningAlgorithm, request.ClientSecret, request.Organization, request.Nonce).ConfigureAwait(false);
```
nandan-bhat, line 223: This change looks correct for MRRT. A refresh that targets a different `audience` and drops `openid` will not return an `id_token`, so skipping validation when the token is absent is the right call and it matches the other flows.

One small thing worth noting: when a caller does request `openid` and expects an `id_token` back, an absent token now passes silently and `IdToken` will be `null` instead of throwing. Not a blocker, but it might be worth a line in the docs telling `openid` callers to null check `response.IdToken`.

Examples.md
kailash-b: Updated the example

src/Auth0.AuthenticationApi/Models/AccessTokenResponse.cs
kailash-b: Added a callout in the PR description. Will ensure to add a callout in the `CHANGELOG` on Release

nandan-bhat (approved):

PR #1057, opened 2026-07-15
feat: Adds Custom Token Exchange support

kailash-b:
### Changes

Adds raw support for **Custom Token Exchange (CTE)** — the OAuth 2.0 Token Exchange grant (RFC 8693) — to the `Auth0.AuthenticationApi` package. A single new grant covers both documented CTE use cases, since they are the same wire call with different inputs:

- **Access token exchange:** exchange a `subject_token` for a normal access token, optionally carrying an `act` (actor) claim.
- **Session Transfer Token (STT):** the same call with `Audience` set to the session-transfer audience and an actor supplied; the response is a Session Transfer Token.

`auth0.net` is a thin, low-level HTTP client, so this adds exactly one grant and no stateful/session/redirect machinery.

**Classes and methods added/changed:**

- **New** `Auth0.AuthenticationApi.Models.TokenExchangeTokenRequest` — request model implementing `IClientAuthentication`. Properties: `SubjectToken`, `SubjectTokenType`, `ActorToken`, `ActorTokenType`, `Audience`, `Scope`, `Reason`, `Organization`, `Nonce`, plus the standard client-auth members.
- **New** `Auth0.AuthenticationApi.Models.TokenType` — static class of well-known token-type URN string constants (`IdToken`, `AccessToken`, `Jwt`, `SessionTransferToken`). Properties stay `string` so callers can also pass custom/future URNs.
- **New** `AuthenticationApiClient.GetTokenAsync(TokenExchangeTokenRequest, CancellationToken)` and the matching `IAuthenticationApiClient` interface method. Builds the `token-exchange` grant body, applies client authentication, and validates the ID token when one is returned (`AssertIdTokenValidIfExisting`). 
- **Changed** `AccessTokenResponse` — added `IssuedTokenType` (`issued_token_type`), the response field indicating which kind of token was issued. `null` for flows that don't return it.

**Usage:**

```csharp
// exchange a subject token for an access token
var tokenResponse = await auth.GetTokenAsync(new TokenExchangeTokenRequest
{
    ClientId = "YOUR_CLIENT_ID",
    ClientSecret = "YOUR_CLIENT_SECRET",
    SubjectToken = "THE_TOKEN_TO_EXCHANGE",
    SubjectTokenType = TokenType.AccessToken,
    Audience = "https://api.example.com",   // optional
    Scope = "read:data"                      // optional
});
Console.WriteLine(tokenResponse.IssuedTokenType);

// Request a Session Transfer Token
var stt = await auth.GetTokenAsync(new TokenExchangeTokenRequest
{
    ClientId = "YOUR_CLIENT_ID",
    ClientSecret = "YOUR_CLIENT_SECRET",
    SubjectToken = "THE_SUBJECT_TOKEN",
    SubjectTokenType = TokenType.AccessToken,
    ActorToken = "THE_ACTOR_TOKEN",
    ActorTokenType = TokenType.AccessToken,
    Audience = "urn:YOUR_AUTH0_DOMAIN:session_transfer",
    Reason = "support-impersonation"
});
```

An `Examples.md` section ("6. Custom Token Exchange (CTE)") documents both scenarios.

### Testing

- [x] This change has been tested on the latest version of the platform/language or why not — verified via mocked-handler integration tests; not exercised against a live Auth0 tenant

### Checklist

### DIFF
--- a/Examples.md
+++ b/Examples.md
@@ -18,6 +18,7 @@
   - [4.1. Pushed Authorization Request (PAR) with authorization details](#41-pushed-authorization-request-par-with-authorization-details)
   - [4.2. Client Initiated Backchannel Authorization (CIBA) with authorization details](#42-client-initiated-backchannel-authorization-ciba-with-authorization-details)
 - [5. Multi-Resource Refresh Token (MRRT)](#5-multi-resource-refresh-token-mrrt)
+- [6. Custom Token Exchange (CTE)](#6-custom-token-exchange-cte)
 
 ## 1. Client Initialization
 
@@ -299,6 +300,63 @@ if (missingScopes.Count > 0)
     Console.WriteLine($"Warning: not all requested scopes were granted. Missing: {string.Join(" ", missingScopes)}");
 }
 ```
+
+[Go to Top](#)
+
+## 6. Custom Token Exchange (CTE)
+
+Custom Token Exchange uses the OAuth 2.0 Token Exchange grant (RFC 8693) to exchange an
+existing token for a new Auth0-issued token. The same call supports two use cases,
+selected by the parameters you pass.
+
+### Phase 1 — exchange a subject token for an access token
+
+```csharp
+using Auth0.AuthenticationApi;
+using Auth0.AuthenticationApi.Models;
+
+var auth = new AuthenticationApiClient("YOUR_AUTH0_DOMAIN");
+
+var tokenResponse = await auth.GetTokenAsync(new TokenExchangeTokenRequest
+{
+    ClientId = "YOUR_CLIENT_ID",
+    ClientSecret = "YOUR_CLIENT_SECRET",
+    SubjectToken = "THE_TOKEN_TO_EXCHANGE",
+    SubjectTokenType = TokenType.AccessToken,   // or a custom URN
+    Audience = "https://api.example.com",       // optional target API
+    Scope = "read:data"                          // optional
+});
+
+Console.WriteLine($"Access Token     : {tokenResponse.AccessToken}");
+Console.WriteLine($"Issued Token Type: {tokenResponse.IssuedTokenType}");
+```
+
+### Phase 2 — request a Session Transfer Token
+
+Set `Audience` to your tenant's session-transfer audience and supply an actor token. The
+response's `IssuedTokenType` will be `TokenType.SessionTransferToken`.
+
+```csharp
+var sttResponse = await auth.GetTokenAsync(new TokenExchangeTokenRequest
+{
+    ClientId = "YOUR_CLIENT_ID",
+    ClientSecret = "YOUR_CLIENT_SECRET",
+    SubjectToken = "THE_SUBJECT_TOKEN",
+    SubjectTokenType = TokenType.AccessToken,
+    ActorToken = "THE_ACTOR_TOKEN",
+    ActorTokenType = TokenType.AccessToken,
+    Audience = "urn:YOUR_AUTH0_DOMAIN:session_transfer",
+    Reason = "support-impersonation"             // optional audit string
+});
+
+if (sttResponse.IssuedTokenType == TokenType.SessionTransferToken)
+{
+    Console.WriteLine($"Session Transfer Token: {sttResponse.AccessToken}");
+}
+```
+
+> `ActorToken` and `ActorTokenType` are both-or-neither — supplying only one throws
+> `ArgumentException` before any network call.
 
 [Go to Top](#)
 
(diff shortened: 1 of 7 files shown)


src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
```diff
@@ -225,6 +225,44 @@ public async Task<AccessTokenResponse> GetTokenAsync(RefreshTokenRequest request
...
+            { "subject_token_type", request.SubjectTokenType }
+        };
+
+        ApplyClientAuthentication(request, body);
+
+        body.AddIfNotEmpty("actor_token", request.ActorToken);
+        body.AddIfNotEmpty("actor_token_type", request.ActorTokenType);
+        body.AddIfNotEmpty("audience", request.Audience);
+        body.AddIfNotEmpty("scope", request.Scope);
+        body.AddIfNotEmpty("reason", request.Reason);
```
nandan-bhat: I could not find `reason` in the Custom Token Exchange docs, and none of the other Auth0 SDKs (node-auth0, auth0-auth-js, nextjs-auth0) send it. The token endpoint just ignores parameters it does not know about, so this value is likely dropped on the server and has no real effect.

The model and the example describe it as an audit string, which is a bit misleading if the server never sees it. Can we confirm with the platform team whether `reason` is a real parameter? If it is not, I think we should drop `Reason` from the request model, this method, and the docs.

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
```diff
@@ -225,6 +225,44 @@ public async Task<AccessTokenResponse> GetTokenAsync(RefreshTokenRequest request
...
+        if (string.IsNullOrEmpty(request.ActorToken) != string.IsNullOrEmpty(request.ActorTokenType))
+            throw new ArgumentException(
+                "ActorToken and ActorTokenType must both be provided together, or both omitted.",
+                nameof(request.ActorToken));
+
+        var body = new Dictionary<string, string> {
+            { "grant_type", "urn:ietf:params:oauth:grant-type:token-exchange" },
+            { "client_id", request.ClientId },
+            { "subject_token", request.SubjectToken },
+            { "subject_token_type", request.SubjectTokenType }
```
nandan-bhat, line 250: `SubjectToken` and `SubjectTokenType` are both required, but we add them straight to the body with no check. If a caller leaves one empty we send an empty field and get back an opaque 400 from the server.

We already do a nice early `ArgumentException` for the actor pair a few lines above, so it feels inconsistent not to guard these two the same way. Can we add a null or empty check for `SubjectToken` and `SubjectTokenType` that throws before the network call?

Examples.md
```diff
@@ -302,6 +303,63 @@ if (missingScopes.Count > 0)
...
+```csharp
+var sttResponse = await auth.GetTokenAsync(new TokenExchangeTokenRequest
+{
+    ClientId = "YOUR_CLIENT_ID",
+    ClientSecret = "YOUR_CLIENT_SECRET",
+    SubjectToken = "THE_SUBJECT_TOKEN",
+    SubjectTokenType = TokenType.AccessToken,
+    ActorToken = "THE_ACTOR_TOKEN",
+    ActorTokenType = TokenType.AccessToken,
+    Audience = "urn:YOUR_AUTH0_DOMAIN:session_transfer",
```
nandan-bhat: Want to double check the Session Transfer Token path here. The Native to Web SSO docs describe getting an STT by exchanging a `refresh_token` (grant_type `refresh_token`) with `audience` set to the session transfer audience. This example instead uses the token exchange grant with an access token as the `subject_token`.

Since this was only tested with mocked handlers, can we verify against a real tenant that the token exchange grant actually returns an STT this way? If the real flow is the refresh token grant, this example may belong on `RefreshTokenRequest` instead. Also worth confirming the exact `subject_token_type` the server expects for STT.

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
kailash-b: Addressed in [commit](55f00482e7cf8483bb7af4e86bb6b424efb44533)

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
kailash-b: Addressed in 55f00482e7cf8483bb7af4e86bb6b424efb44533

Examples.md
kailash-b: Addressed in 55f00482e7cf8483bb7af4e86bb6b424efb44533. 
We can use access token as a `subject_token`, however, we cannot use the `subject_token_type` as the reserved `urn:ietf:params:oauth:token-type:access_token`. 
We have corrected this in the mock test cases and in the examples.

nandan-bhat (approved):

PR #1064, opened 2026-07-17
feat: Adds Token Vault support

kailash-b:
### Changes

Adds **Token Vault** support to the `Auth0.AuthenticationApi` package. Token Vault lets a private client exchange an Auth0 token for an access token issued by one of Auth0's *federated connections* (Google, Facebook, etc.), so the app can call that provider's APIs on the user's behalf. 

**Classes/methods added:**

- **New model** `FederatedConnectionAccessTokenRequest` (implements `IClientAuthentication`) — carries `SubjectToken`, `SubjectTokenType`, `Connection` (required), optional `LoginHint`, `Nonce`, and the standard client-auth members (`ClientId`, `ClientSecret`, `ClientAssertionSecurityKey`, `ClientAssertionSecurityKeyAlgorithm`, `SigningAlgorithm`).
- **New overload** `Task<AccessTokenResponse> GetTokenAsync(FederatedConnectionAccessTokenRequest request, CancellationToken)` on `IAuthenticationApiClient` / `AuthenticationApiClient`. Builds the request body, applies client authentication, hardcodes `requested_token_type`, POSTs to `/oauth/token`, and reuses the existing `AccessTokenResponse` (the federated token is returned in `AccessToken`; `issued_token_type` is surfaced via `IssuedTokenType`).
- **New `TokenType` constants**: `RefreshToken` (`urn:ietf:params:oauth:token-type:refresh_token`) and `FederatedConnectionAccessToken` (`http://auth0.com/oauth/token-type/federated-connection-access-token`).

**Usage:**

```csharp
var auth = new AuthenticationApiClient("YOUR_AUTH0_DOMAIN");

var tokenResponse = await auth.GetTokenAsync(new FederatedConnectionAccessTokenRequest
{
    ClientId = "YOUR_CLIENT_ID",
    ClientSecret = "YOUR_CLIENT_SECRET",
    SubjectToken = "THE_AUTH0_ACCESS_TOKEN",
    SubjectTokenType = TokenType.AccessToken,
    Connection = "google-oauth2",
    LoginHint = "THE_GOOGLE_USER_ID"   // optional
});

Console.WriteLine($"Federated Access Token: {tokenResponse.AccessToken}");
```

### References

- Auth0 docs: [Access token exchange with Token Vault](https://auth0.com/docs/secure/call-apis-on-users-behalf/token-vault/access-token-exchange-with-token-vault)

### Testing

### Checklist

### DIFF
--- a/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
+++ b/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
@@ -275,6 +275,16 @@ public class AuthenticationApiClient : IAuthenticationApiClient
     {
         request.ThrowIfNull();
 
+        if (string.IsNullOrEmpty(request.SubjectToken))
+            throw new ArgumentException(
+                "SubjectToken is required for a federated connection access token exchange.",
+                nameof(request.SubjectToken));
+
+        if (string.IsNullOrEmpty(request.SubjectTokenType))
+            throw new ArgumentException(
+                "SubjectTokenType is required for a federated connection access token exchange.",
+                nameof(request.SubjectTokenType));
+
         if (string.IsNullOrEmpty(request.Connection))
             throw new ArgumentException(
                 "Connection is required for a federated connection access token exchange.",
--- a/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
+++ b/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
@@ -271,6 +271,41 @@ public class AuthenticationApiClient : IAuthenticationApiClient
     }
 
     /// <inheritdoc/>
+    public async Task<AccessTokenResponse> GetTokenAsync(FederatedConnectionAccessTokenRequest request, CancellationToken cancellationToken = default)
+    {
+        request.ThrowIfNull();
+
+        if (string.IsNullOrEmpty(request.Connection))
+            throw new ArgumentException(
+                "Connection is required for a federated connection access token exchange.",
+                nameof(request.Connection));
+
+        var body = new Dictionary<string, string> {
+            { "grant_type", "urn:auth0:params:oauth:grant-type:token-exchange:federated-connection-access-token" },
+            { "client_id", request.ClientId },
+            { "subject_token", request.SubjectToken },
+            { "subject_token_type", request.SubjectTokenType },
+            { "requested_token_type", TokenType.FederatedConnectionAccessToken },
+            { "connection", request.Connection }
+        };
+
+        ApplyClientAuthentication(request, body);
+
+        body.AddIfNotEmpty("login_hint", request.LoginHint);
+
+        var response = await connection.SendAsync<AccessTokenResponse>(
+            HttpMethod.Post,
+            tokenUri,
+            body,
+            cancellationToken: cancellationToken
+        ).ConfigureAwait(false);
+
+        await AssertIdTokenValidIfExisting(response.IdToken, request.ClientId, request.SigningAlgorithm, request.ClientSecret, null, request.Nonce).ConfigureAwait(false);
+
+        return response;
+    }
+
+    /// <inheritdoc/>
     public async Task<AccessTokenResponse> GetTokenAsync(ResourceOwnerTokenRequest request, CancellationToken cancellationToken = default)
     {
         request.ThrowIfNull();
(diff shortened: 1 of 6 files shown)


src/Auth0.AuthenticationApi/Models/FederatedConnectionAccessTokenRequest.cs
```diff
@@ -0,0 +1,74 @@
...
+/// </summary>
+/// <remarks>
+/// The client must be a private client. Today the subject token must be an Auth0 refresh
+/// token (<see cref="TokenType.RefreshToken"/>); support for access-token subjects is
+/// planned by Auth0.
+/// </remarks>
+public class FederatedConnectionAccessTokenRequest : IClientAuthentication
+{
+    /// <summary>
+    /// The Auth0 token being exchanged. Currently this must be a valid Auth0 refresh token.
```
nandan-bhat: This doc says the subject token must be an Auth0 `refresh token`, but the example, every test in this PR, and the Token Vault docs all use an `access token` (`subject_token_type` = `urn:ietf:params:oauth:token-type:access_token`). The Auth0 docs describe `subject_token` as the Auth0 access token that identifies the user.

The class remark on line 11 and the `SubjectTokenType` note on line 25 say the same refresh token thing, so all three spots contradict what the code actually does. Since this text shows up in IntelliSense, a user following it would pass a refresh token and get a 400. Can we update these to say `access token` (`TokenType.AccessToken`)? If refresh tokens are also valid, then let us word it as access or refresh token, but right now nothing exercises the refresh token path.

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
```diff
@@ -270,6 +270,41 @@ public async Task<AccessTokenResponse> GetTokenAsync(TokenExchangeTokenRequest r
...
+        if (string.IsNullOrEmpty(request.Connection))
+            throw new ArgumentException(
+                "Connection is required for a federated connection access token exchange.",
+                nameof(request.Connection));
+
+        var body = new Dictionary<string, string> {
+            { "grant_type", "urn:auth0:params:oauth:grant-type:token-exchange:federated-connection-access-token" },
+            { "client_id", request.ClientId },
+            { "subject_token", request.SubjectToken },
+            { "subject_token_type", request.SubjectTokenType },
```
nandan-bhat, line 297: `SubjectToken` and `SubjectTokenType` are both required but we add them to the body with no check. We already throw a clear `ArgumentException` for a missing `Connection` a few lines up, so it feels odd to let these two through and get back an opaque 400 from the server.

Can we add the same kind of early null or empty guard for `SubjectToken` and `SubjectTokenType`?

src/Auth0.AuthenticationApi/Models/TokenType.cs
```diff
@@ -30,4 +30,14 @@ public static class TokenType
     /// Indicates that the token is an Auth0 Session Transfer Token: <c>urn:auth0:params:oauth:token-type:session_transfer_token</c>.
     /// </summary>
     public const string SessionTransferToken = "urn:auth0:params:oauth:token-type:session_transfer_token";
+
+    /// <summary>
+    /// Indicates that the token is an OAuth 2.0 refresh token: <c>urn:ietf:params:oauth:token-type:refresh_token</c>.
+    /// </summary>
+    public const string RefreshToken = "urn:ietf:params:oauth:token-type:refresh_token";
```
nandan-bhat, line 37: This `RefreshToken` constant does not seem to be used anywhere in the feature. All the code paths and tests use `AccessToken`. It looks like it was added only to support the refresh token wording in the request model, which I think is not correct (left a note there).

If refresh token subjects are genuinely supported, can we add a test that uses this? Otherwise I would drop it so we do not signal a path that does not work. Low priority either way.

src/Auth0.AuthenticationApi/Models/FederatedConnectionAccessTokenRequest.cs
kailash-b: Addressed in 5b53c2945094552e143920fbaf41f3f059bad7e4
`subject_token_type` can be both `AccessToken` or `RefreshToken`. Updated the code docs to reflect the same.

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
kailash-b: Addressed in 5b53c2945094552e143920fbaf41f3f059bad7e4
Added validation

src/Auth0.AuthenticationApi/Models/TokenType.cs
kailash-b: Addressed in 5b53c2945094552e143920fbaf41f3f059bad7e4
This is a public property. Also, now we have added test cases that use this `RefreshToken` as well.

nandan-bhat (approved):

PR #1094, opened 2026-08-13
feat: Adds On-Behalf of Token Exchange support

kailash-b:
### Changes

This PR adds **On-Behalf-Of (OBO) Token Exchange** support to the Authentication API client, letting a service (for example, an MCP server) exchange an incoming user access token for a short-lived, audience-scoped access token that preserves the user's identity and actor attribution.

- Adds a new `GetTokenOnBehalfOfAsync` method to the Authentication API client that performs the exchange. The caller supplies the incoming user token, the target audience, and optional scope/organization; the grant type and subject-token type are set internally.
- The response exposes helpers to read the actor information from the exchanged token: one for the **current actor** (the value to use for authorization decisions) and one for the **full delegation chain** (for audit/logging only).
- Token validation is out of scope 
- Documentation and a usage example have been added to `README.md` and `Examples.md`.

### Testing
- Unit tests cover the exchange request (correct parameters sent), input validation, response mapping, and the actor helpers. 

### Checklist

### DIFF
--- a/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
+++ b/src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
@@ -271,6 +271,44 @@ public class AuthenticationApiClient : IAuthenticationApiClient
     }
 
     /// <inheritdoc/>
+    public async Task<OnBehalfOfTokenResponse> GetTokenOnBehalfOfAsync(OnBehalfOfTokenRequest request, CancellationToken cancellationToken = default)
+    {
+        request.ThrowIfNull();
+
+        if (string.IsNullOrEmpty(request.SubjectToken))
+            throw new ArgumentException(
+                "SubjectToken is required.", nameof(request.SubjectToken));
+
+        if (string.IsNullOrEmpty(request.Audience))
+            throw new ArgumentException(
+                "Audience is required.", nameof(request.Audience));
+
+        var tokenExchangeRequest = new TokenExchangeTokenRequest
+        {
+            SubjectToken = request.SubjectToken,
+            SubjectTokenType = TokenType.AccessToken,
+            Audience = request.Audience,
+            Scope = request.Scope,
+            Organization = request.Organization,
+            ClientId = request.ClientId,
+            ClientSecret = request.ClientSecret,
+            ClientAssertionSecurityKey = request.ClientAssertionSecurityKey,
+            ClientAssertionSecurityKeyAlgorithm = request.ClientAssertionSecurityKeyAlgorithm
+        };
+
+        var response = await GetTokenAsync(tokenExchangeRequest, cancellationToken).ConfigureAwait(false);
+
+        return new OnBehalfOfTokenResponse
+        {
+            AccessToken = response.AccessToken,
+            TokenType = response.TokenType,
+            ExpiresIn = response.ExpiresIn,
+            Scope = response.Scope,
+            IssuedTokenType = response.IssuedTokenType
+        };
+    }
+
+    /// <inheritdoc/>
     public async Task<AccessTokenResponse> GetTokenAsync(FederatedConnectionAccessTokenRequest request, CancellationToken cancellationToken = default)
     {
         request.ThrowIfNull();
--- a/src/Auth0.AuthenticationApi/Models/Actor.cs
+++ b/src/Auth0.AuthenticationApi/Models/Actor.cs
@@ -14,15 +14,26 @@ namespace Auth0.AuthenticationApi.Models;
 public class Actor
 {
     /// <summary>
+    /// Initializes a new <see cref="Actor"/>. Used by the deserializer to populate the
+    /// read-only properties; the type is otherwise only produced by decoding an access token.
+    /// </summary>
+    [JsonConstructor]
+    public Actor(string? subject, Actor? act)
+    {
+        Subject = subject;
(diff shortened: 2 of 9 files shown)


src/Auth0.AuthenticationApi/Models/OnBehalfOfTokenResponse.cs
```diff
@@ -0,0 +1,115 @@
...
+        // Volatile.Read/Write publish the immutable holder with acquire/release semantics so
+        // a concurrent reader that sees the reference also sees its fully-constructed fields
+        // (a plain write gives no such guarantee on weak memory models such as ARM64).
+        var token = AccessToken;
+        var cache = Volatile.Read(ref actorCache);
+
+        if (cache is null || !ReferenceEquals(cache.Token, token))
+        {
+            cache = new ActorCache(token, ActClaimReader.ReadActor(token));
+            Volatile.Write(ref actorCache, cache);
```
nandan-bhat, line 93: There is a write-write race here when `AccessToken` is reassigned on one thread while another thread is inside `GetActor()`.

Scenario: thread A snapshots `token = "v1"` and is about to write the cache. Thread B reassigns `AccessToken = "v2"`, computes `ActorCache("v2", actor_v2)`, and writes it. Then thread A resumes and overwrites with `ActorCache("v1", actor_v1)`. The cache now holds the actor for the old token. A third concurrent caller will read the wrong actor for `"v2"` until the cache is invalidated on the next call.

The fix is to replace `Volatile.Write` with `Interlocked.CompareExchange(ref actorCache, cache, null)` - or a snapshot-based CAS - so only the first writer wins and a stale overwrite is silently discarded. The concurrent-read test covers a stable token, but there is no test for this concurrent-mutation path.

src/Auth0.AuthenticationApi/Models/Actor.cs
```diff
@@ -0,0 +1,28 @@
...
+/// (audit/logging).
+/// </remarks>
+public class Actor
+{
+    /// <summary>
+    /// The <c>sub</c> of this actor. May be <c>null</c> if the decoded <c>act</c> claim did
+    /// not contain a <c>sub</c>.
+    /// </summary>
+    [JsonPropertyName("sub")]
+    public string? Subject { get; set; }
```
nandan-bhat: `Subject` and `Act` are both settable (`{ get; set; }`). `GetDelegationChain()` returns the same `Actor` instance that is held inside the cached `ActorCache`, so a caller can do `response.GetDelegationChain()!.Subject = "something"` and permanently corrupt the cached actor. The next call to `GetCurrentActor()` will then return the wrong value.

Since `Actor` is only ever constructed by JSON deserialization or by `ActClaimReader`, changing both properties to `{ get; init; }` would make the type immutable without breaking anything. That way the cached instance is safe to hand out directly.

src/Auth0.AuthenticationApi/AuthenticationApiClient.cs
```diff
@@ -270,6 +270,44 @@ public async Task<AccessTokenResponse> GetTokenAsync(TokenExchangeTokenRequest r
...
+
+        var response = await GetTokenAsync(tokenExchangeRequest, cancellationToken).ConfigureAwait(false);
+
+        return new OnBehalfOfTokenResponse
+        {
+            AccessToken = response.AccessToken,
+            TokenType = response.TokenType,
+            ExpiresIn = response.ExpiresIn,
+            Scope = response.Scope,
+            IssuedTokenType = response.IssuedTokenType
```
nandan-bhat, line 307: `RefreshToken` and `IdToken` from the inner `AccessTokenResponse` are not mapped here and are silently dropped. The PR description says OBO issues no refresh or ID token, which is correct today, so this is fine in practice.

Just worth a note: if the server ever returns an `id_token` in this flow, it will actually be validated inside the `GetTokenAsync(TokenExchangeTokenRequest)` call above (via `AssertIdTokenValidIfExisting`), but the caller will never see it because `OnBehalfOfTokenResponse` has no `IdToken` property. A quick comment or a doc note on the response class saying these fields are intentionally excluded would help future maintainers understand this was a deliberate choice and not an oversight.

kailash-b: This is by design. The OBO flow will not issue `id_token` or `refresh_token` today and if at it does, there should be a conscious change from our side on the SDK to expose it rather expose it automatically.

src/Auth0.AuthenticationApi/Models/OnBehalfOfTokenResponse.cs
kailash-b: The scenario is real, it is near impossible to occur given that this method is on a token response. 
Also, there will be no scenario where a wrong actor can be read since we are checking that the token in the cache is matching the token in the request. The worst possible outcome is that there could be a recompute of the value for the token that gets over-ridden. 
No changes required.

src/Auth0.AuthenticationApi/Models/Actor.cs
kailash-b: Ack. 
Addressed in 32485ba816b7b7464d7a413c370011b832dac81c. 
We have exposed a constructor and made the properties read-only as highlighted.

nandan-bhat (approved):

